PR #168, opened 2025-09-10
[fix][coze-loop]: resolve port conflict with coze-studio

shx815:
#### What type of PR is this?
fix

#### Check the PR title.

- [x] This PR title match the format:[fix][coze-loop]: resolve port conflict with coze-studio

#### (Optional) Translate the PR title into Chinese.
[修复][coze-loop] 将端口8888改为8880以解决冲突

**en:**  
- **Changed files:**  
  `release/deployment/docker-compose/.env`: Modified `COZE_LOOP_APP_OPENAPI_PORT` from `8888` to `8880`  

- **Reason for change:**  
  Resolves port conflict between Coze Loop and Coze Studio, which both default to `8888`.  

- **Testing performed:**  
  - Verified local deployment with `docker compose up -d`  
  - Confirmed no port binding errors in container logs  
  - Validated Coze Loop access via `http://localhost:8880`  

**zh(optional):**  
- **修改文件:**  
  `.env` 文件：将 `COZE_LOOP_APP_OPENAPI_PORT` 从 `8888` 修改为 `8880`  

- **变更原因:**  
  解决 Coze Loop 与 Coze Studio 默认端口 `8888` 的冲突问题  

- **测试情况:**  
  - 本地部署 `docker compose up -d` 无端口绑定错误  
  - 容器日志无异常输出  
  - 通过 `http://localhost:8880` 可正常访问 Coze Loop


release/deployment/docker-compose/docker-compose.yml
```diff
@@ -8,7 +8,7 @@ services:
     networks:
       - coze-loop-network
     ports:
-      - "${COZE_LOOP_APP_OPENAPI_PORT}:8888"
```
lsy357: The container port is still the 8888 port that Hertz listens on, so directly modifying it will cause problems.

shx815: Sorry for my mistake, I have recovered the container port to 8888.

release/deployment/docker-compose/.env
```diff
@@ -3,7 +3,7 @@ COZE_LOOP_APP_IMAGE_REGISTRY=docker.io
 COZE_LOOP_APP_IMAGE_REPOSITORY=cozedev
 COZE_LOOP_APP_IMAGE_NAME=coze-loop
 COZE_LOOP_APP_IMAGE_TAG=1.1.0
-COZE_LOOP_APP_OPENAPI_PORT=8888
+COZE_LOOP_APP_OPENAPI_PORT=8880
```
lsy357, line 6: Hi, I think the PORT here can't be changed. Because after modifying it and redeploying, the SDK client code also needs to be updated to listen on the new port. This incompatible change requires careful handling.

lsy357: In your case, changing the port may need to be addressed by automatically mounting customized configurations.

lsy357: cc @kidkidkid

shx815: Sorry I took it to easy.
 But i think it will make sense if we resolve port conflict with coze-studio, since there is need to use both of them.

lsy357: We will carefully consider and resolve the issue you encountered during using. We will continue to communicate with you once we have a solution.

PR #182, opened 2025-09-15
[fix][infra] Fix nginx data volume name

freeme:
Fixed the environment variable `COZE_LOOP_NGINX_DATA_VOLUME_NAME` in .env file to ensure proper volume mounting


release/deployment/docker-compose/.env
```diff
@@ -65,4 +65,4 @@ COZE_LOOP_NGINX_IMAGE_REPOSITORY=library
 COZE_LOOP_NGINX_IMAGE_NAME=nginx
 COZE_LOOP_NGINX_IMAGE_TAG=1.28.0
 COZE_LOOP_NGINX_PORT=8082
-COZE_LOOP_NGINX_DATA_VOLUME_NAME=coze-loop-nginx-data
```
lsy357, line 68: Thanks, I think it would be better to rm the default value of COZE_LOOP_NGINX_DATA_VOLUME_NAME variable in docker-compose.yml

freeme:
Fixed in commit: b7544101ada59c716a20fdd646e500ab7ca2f066

release/deployment/docker-compose/.env
lsy357, line 68: Keep the name coze-loop-nginx-data?

lsy357: I will modify here based on your code.

lsy357 (approved):

PR #191, opened 2025-09-17
Feat/insight analysis

hanxu-111:
#### Check the PR title.

en:
zh(optional):


backend/modules/evaluation/domain/component/rpc/notify.go
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) 2025 coze-dev Authors
+// SPDX-License-Identifier: Apache-2.0
+
+package rpc
+
+import "context"
+
+//go:generate mockgen -destination=mocks/notify.go -package=mocks . INotifyRPCAdapter
+type INotifyRPCAdapter interface {
+	SendLarkMessageCard(ctx context.Context, userID, cardID string, param map[string]string) error
```
HymanShi: 这里是不是不要带上lark比较好

backend/modules/evaluation/domain/service/insight_analysis_impl.go
```diff
@@ -0,0 +1,334 @@
...
+	if analysisRecord.Status == entity.InsightAnalysisStatus_Running ||
+		analysisRecord.Status == entity.InsightAnalysisStatus_Failed {
+		return analysisRecord, nil
+	}
+
+	report, _, err := e.agentAdapter.GetReport(ctx, spaceID, ptr.From(analysisRecord.AnalysisReportID))
+	if err != nil {
+		return nil, err
+	}
+	// 聚合报告生成状态
```
HymanShi: 这块代码还需要吗

backend/modules/evaluation/domain/service/insight_analysis_impl.go
```diff
@@ -0,0 +1,334 @@
...
+	}
+	logs.CtxInfo(ctx, "notifyAnalysisComplete userInfos: %v", userInfos)
+
+	if len(userInfos) != 1 || userInfos[0] == nil {
+		return nil
+	}
+
+	userInfo := userInfos[0]
+	logs.CtxInfo(ctx, "notifyAnalysisComplete userInfo: %v", userInfo)
+	const cardID = "AAq9DvIYd2qHu"
```
HymanShi: 这个参数可以隐藏到infra的实现里吗

hanxu-111: 考虑后续其他场景需要消息通知的话，还是需要在业务层传入不同的cardID模板的

backend/modules/evaluation/domain/service/insight_analysis_impl.go
```diff
@@ -0,0 +1,326 @@
...
+			ExptID:             exptID,
+			ExptResultFilePath: ptr.Of(exptResultFilePath),
+			AnalysisReportID:   ptr.Of(analysisReportID),
+			Status:             entity.InsightAnalysisStatus_Running,
+		}
+		if err != nil {
+			record.Status = entity.InsightAnalysisStatus_Failed
+			// 不重试
+			err = nil
+		}
```
CozeLoop, line 104: **🚫Must-Fix: Critical Error Handling Issue in Defer Function**

**Issue Description**: The defer function modifies the named return variable `err` by setting it to `nil` on line 104, which silently swallows errors and prevents proper error propagation to the caller. This can lead to silent failures and make debugging extremely difficult.

**Severity Level**: Critical

**Problems**:
1. `err = nil` on line 104 masks actual errors that occurred during execution
2. The caller will receive `nil` error even when the operation failed
3. This violates Go's explicit error handling principle
4. Makes the system unreliable and hard to debug

**Improvement Example**:
```go
defer func() {
    record := &entity.ExptInsightAnalysisRecord{
        ID:                 recordID,
        SpaceID:            spaceID,
        ExptID:             exptID,
        ExptResultFilePath: ptr.Of(exptResultFilePath),
        AnalysisReportID:   ptr.Of(analysisReportID),
        Status:             entity.InsightAnalysisStatus_Running,
    }
    if err != nil {
        record.Status = entity.InsightAnalysisStatus_Failed
        // Remove this line: err = nil
    }
    if updateErr := e.repo.UpdateAnalysisRecord(ctx, record); updateErr != nil {
        logs.CtxError(ctx, "UpdateAnalysisRecord failed: %v", updateErr)
        // Consider whether to override err with updateErr if err is nil
        if err == nil {
            err = updateErr
        }
    }
}()
```

backend/modules/evaluation/application/experiment_app.go
```diff
@@ -1077,3 +1080,196 @@ func (e *experimentApplication) GetExptResultExportRecord(ctx context.Context, r
...
+	if err != nil {
+		return nil, err
+	}
+	return &expt.DeleteExptInsightAnalysisRecordResponse{
+		BaseResp: base.NewBaseResp(),
+	}, nil
+}
+
+func (e *experimentApplication) GetExptInsightAnalysisRecord(ctx context.Context, req *expt.GetExptInsightAnalysisRecordRequest) (r *expt.GetExptInsightAnalysisRecordResponse, err error) {
+	session := &entity.Session{UserID: strconv.FormatInt(req.GetSession().GetUserID(), 10)}
```
CozeLoop: **🚫Must-Fix: Potential Null Pointer Dereference**

**Issue Description**: Direct access to `req.GetSession().GetUserID()` without null checks can cause panic if `req.GetSession()` returns nil. This pattern is inconsistent with other methods in the same file that properly handle nil sessions.

**Severity Level**: Critical

**Problems**:
1. No null check for `req.GetSession()` before calling `GetUserID()`
2. Inconsistent session handling compared to other methods
3. Can cause runtime panic and service crash

**Improvement Example**:
```go
session := entity.NewSession(ctx)
if req.Session != nil && req.Session.UserID != nil {
    session = &entity.Session{
        UserID: strconv.FormatInt(gptr.Indirect(req.Session.UserID), 10),
    }
}
```

**Reference**: This same pattern is correctly implemented in `InsightAnalysisExperiment`, `ListExptInsightAnalysisRecord`, and other methods in this file.

backend/modules/evaluation/infra/rpc/agent/agent.go
```diff
@@ -0,0 +1,25 @@
...
+	"github.com/coze-dev/coze-loop/backend/modules/evaluation/domain/entity"
+)
+
+type AgentAdapter struct{}
+
+func NewAgentAdapter() rpc.IAgentAdapter {
+	return &AgentAdapter{}
+}
+
+func (a AgentAdapter) CallTraceAgent(ctx context.Context, spaceID int64, url string) (int64, error) {
```
CozeLoop, line 21: **🚨Should-Fix: Stub Implementation in Production Code**

**Issue Description**: The `CallTraceAgent` method returns hardcoded zero values (`return 0, nil`), which suggests this is a stub implementation. This will cause the insight analysis feature to not work correctly in production.

**Severity Level**: Medium

**Problems**:
1. Always returns `0` as report ID, which is likely invalid
2. No actual RPC call to the trace agent service
3. Feature will appear to work but won't generate real analysis reports
4. Could cause confusion and debugging difficulties

**Improvement Example**:
```go
func (a AgentAdapter) CallTraceAgent(ctx context.Context, spaceID int64, url string) (int64, error) {
    // TODO: Implement actual RPC call to trace agent service
    // Example:
    // req := &traceagent.AnalysisRequest{
    //     SpaceID: spaceID,
    //     DataURL: url,
    // }
    // resp, err := a.traceAgentClient.CreateAnalysis(ctx, req)
    // if err != nil {
    //     return 0, err
    // }
    // return resp.ReportID, nil
    
    return 0, errors.New("CallTraceAgent not implemented yet")
}
```

**Note**: Same issue exists in `GetReport` method on line 23.

backend/modules/evaluation/domain/service/expt_export_impl.go
```diff
@@ -492,6 +511,22 @@ func (e *exportCSVHelper) buildRows(ctx context.Context) ([][]string, error) {
...
+			// logID
+			if e.withLogID {
+				logID := ""
+				if payload.SystemInfo != nil {
+					logID = ptr.From(payload.SystemInfo.LogID)
+				}
+				traceID := ""
+				if payload.TargetOutput == nil ||
+					payload.TargetOutput.EvalTargetRecord == nil {
+					traceID = payload.TargetOutput.EvalTargetRecord.TraceID
```
CozeLoop, line 523: **🚫Must-Fix: Logical Error in Null Check**

**Issue Description**: The condition `payload.TargetOutput == nil || payload.TargetOutput.EvalTargetRecord == nil` followed by accessing `payload.TargetOutput.EvalTargetRecord.TraceID` will cause a null pointer dereference when either condition is true.

**Severity Level**: Critical

**Problems**:
1. When `payload.TargetOutput` is `nil`, accessing `payload.TargetOutput.EvalTargetRecord.TraceID` will panic
2. When `payload.TargetOutput.EvalTargetRecord` is `nil`, accessing `.TraceID` will panic
3. The logic should be inverted to safely access the field

**Improvement Example**:
```go
traceID := ""
if payload.TargetOutput != nil &&
    payload.TargetOutput.EvalTargetRecord != nil {
    traceID = payload.TargetOutput.EvalTargetRecord.TraceID
}
```

**Reference**: This follows the same safe null-checking pattern used for `logID` above.

backend/modules/evaluation/infra/mq/rocket/consumer/expt_export.go
```diff
@@ -49,11 +51,20 @@ func (e *ExptExportConsumer) HandleMessage(ctx context.Context, ext *mq.MessageE
...
-	if err != nil {
-		// 不进行重试
-		logs.CtxError(ctx, "ExptExportConsumer DoExportCSV fail, expt_id:%v, err: %v", event.ExperimentID, err)
-		return nil
+	switch event.ExportScene {
+	case entity.ExportSceneInsightAnalysis:
+		err = e.exptInsightAnalysisService.GenAnalysisReport(ctx, event.SpaceID, event.ExperimentID, event.ExportID, event.CreateAt)
+		if err != nil {
+			logs.CtxError(ctx, "ExptExportConsumer GenAnalysisReport fail, expt_id:%v, err: %v", event.ExperimentID, err)
+			return err
```
CozeLoop: **🚨Should-Fix: Inconsistent Error Handling in Message Consumer**

**Issue Description**: There's inconsistent error handling between the insight analysis case and the default case. The insight analysis case returns the error (causing message retry), while the default case returns `nil` (preventing retry).

**Severity Level**: Medium

**Problems**:
1. Insight analysis failures will cause infinite retries if the error is persistent
2. Inconsistent behavior between different export scenarios
3. Could lead to message queue buildup and system instability
4. No clear retry strategy for insight analysis failures

**Improvement Example**:
```go
case entity.ExportSceneInsightAnalysis:
    err = e.exptInsightAnalysisService.GenAnalysisReport(ctx, event.SpaceID, event.ExperimentID, event.ExportID, event.CreateAt)
    if err != nil {
        logs.CtxError(ctx, "ExptExportConsumer GenAnalysisReport fail, expt_id:%v, err: %v", event.ExperimentID, err)
        // Consider whether insight analysis should retry or not
        // If no retry needed: return nil
        // If retry needed: return err
        return nil // or implement proper retry logic with backoff
    }
```

**Recommendation**: Align error handling strategy across both cases based on business requirements.

backend/modules/evaluation/infra/repo/experiment/mysql/expt_insight_analysis_record.go
```diff
@@ -0,0 +1,98 @@
...
+	if err != nil {
+		return nil, 0, err
+	}
+	return finds, total, nil
+}
+
+func (e exptInsightAnalysisRecordDAO) Delete(ctx context.Context, spaceID, exptID, recordID int64) error {
+	po := &model.ExptInsightAnalysisRecord{}
+	db := e.db.NewSession(ctx)
+	err := db.Where("space_id = ? AND expt_id = ?  AND record_id = ?", spaceID, exptID, recordID).
```
CozeLoop: **🚫Must-Fix: Incorrect Column Name in Delete Query**

**Issue Description**: The DELETE query uses `record_id` in the WHERE clause, but based on the model structure and other methods, the correct column name should be `id`. This will cause the delete operation to fail.

**Severity Level**: Critical

**Problems**:
1. Column name mismatch: using `record_id` instead of `id`
2. Delete operations will not work correctly
3. Inconsistent with other methods in the same class that use `id`

**Improvement Example**:
```go
func (e exptInsightAnalysisRecordDAO) Delete(ctx context.Context, spaceID, exptID, recordID int64) error {
    po := &model.ExptInsightAnalysisRecord{}
    db := e.db.NewSession(ctx)
    err := db.Where("space_id = ? AND expt_id = ? AND id = ?", spaceID, exptID, recordID).
        Delete(po).Error
    if err != nil {
        return err
    }
    return nil
}
```

**Reference**: Other methods like `GetByID` and `Update` correctly use `id` as the primary key column name.

CozeLoop (changes requested):
## Code Review Summary

I've conducted a systematic expert-level code review of this insight analysis feature PR. The implementation introduces significant new functionality but contains several **critical issues** that must be addressed before merging.

### Critical Issues Found (🚫Must-Fix):
1. **Error handling bug in defer function** - Silently swallows errors
2. **Null pointer dereference risk** - Inconsistent session handling  
3. **Logical error in null checks** - Will cause runtime panics
4. **Incorrect SQL column name** - Delete operations will fail

### Medium Priority Issues (🚨Should-Fix):
1. **Stub implementation** - Agent adapter returns hardcoded values
2. **Inconsistent error handling** - Message consumer retry logic

These issues could cause production failures, data inconsistencies, and poor user experience. Please address the critical issues before requesting another review.

### Positive Aspects:
- Well-structured domain-driven design
- Comprehensive test coverage for new functionality
- Proper separation of concerns across layers
- Good use of Go interfaces and dependency injection

The overall architecture and design approach is solid, but the implementation details need refinement to meet production quality standards.

CozeLoop (changes requested):
Summary of key findings:

Must-Fix:
1) Voting toggle path (Upvote/Downvote) unconditionally inserts — violates UNIQUE KEY on toggles; implement idempotent upsert or update-if-exists.
2) SQL: expt_insight_analysis_feedback_vote.sql missing trailing semicolon and uses non-standard "Unique KEY" casing.
3) DB migrations only added for Docker Compose; mirror the same DDLs under Helm init-sql for k8s.

Should-Fix:
4) ListExptInsightAnalysisRecord response lacks BaseResp and uses a misleading field name (should be expt_insight_analysis_records, requires IDL/code regen).
5) Event naming: CreateAt -> CreatedAt to align with conventions.

Suggestion:
6) Avoid logging full userInfos/userInfo (PII); lower to debug or log only identifiers.

See inline comments for precise locations and code examples.

backend/modules/evaluation/domain/service/insight_analysis_impl.go
```diff
@@ -0,0 +1,322 @@
...
+		exptResultFilePath string
+		analysisReportID   int64
+	)
+	defer func() {
+		record := &entity.ExptInsightAnalysisRecord{
+			ID:                 recordID,
+			SpaceID:            spaceID,
+			ExptID:             exptID,
+			ExptResultFilePath: ptr.Of(exptResultFilePath),
+			AnalysisReportID:   ptr.Of(analysisReportID),
```
CozeLoop: [Must-Fix] Do not persist AnalysisReportID when it is zero

Problem: In the deferred update, AnalysisReportID is always set via ptr.Of(analysisReportID). If an error occurs before CallTraceAgent succeeds, analysisReportID remains 0, resulting in a non-nil pointer to 0 being stored. Subsequent checks (e.g., checkAnalysisReportGenStatus) will treat the record as having a report ID and will call GetReport(..., 0), which is invalid and leads to incorrect flow handling.

Why this matters: Persisting an invalid report ID makes the state machine inconsistent and can cause repeated failed polling. This violates robustness expectations in error handling (see Effective Go / Go Code Review Comments: keep data consistent on error paths) and internal service code guidelines.

Suggested fix: Only set AnalysisReportID when a valid ID (>0) is obtained. Otherwise, keep it nil.

Example:
```go
defer func() {
    record := &entity.ExptInsightAnalysisRecord{
        ID:                 recordID,
        SpaceID:            spaceID,
        ExptID:             exptID,
        ExptResultFilePath: ptr.Of(exptResultFilePath),
        Status:             entity.InsightAnalysisStatus_Running,
    }
    if analysisReportID > 0 {
        record.AnalysisReportID = ptr.Of(analysisReportID)
    }
    if err != nil {
        record.Status = entity.InsightAnalysisStatus_Failed
    }
    if err1 := e.repo.UpdateAnalysisRecord(ctx, record); err1 != nil {
        logs.CtxError(ctx, "UpdateAnalysisRecord failed: %v", err1)
        if err == nil {
            err = err1
        }
    }
}()
```

This ensures records remain consistent and avoids calling GetReport with an invalid ID.

backend/modules/evaluation/domain/service/insight_analysis_impl.go
```diff
@@ -0,0 +1,322 @@
...
+	}
+
+	var ttl int64 = 24 * 60 * 60
+	signOpt := fileserver.SignWithTTL(time.Duration(ttl) * time.Second)
+
+	url, _, err := e.fileClient.SignDownloadReq(ctx, fileName, signOpt)
+	if err != nil {
+		return err
+	}
+	logs.CtxInfo(ctx, "GenAnalysisReport get csv url=%v", url)
```
CozeLoop: [Should-Fix] Avoid logging pre-signed URLs (sensitive information)

Problem: The INFO log prints the full pre-signed download URL of the CSV file. Pre-signed URLs typically embed short-lived credentials/tokens and should be treated as sensitive. Emitting them at INFO can leak access in logs and observability backends.

Recommendation: Log only the file name or a redacted URL (e.g., host + path without query) and/or hash. Keep detailed URLs at DEBUG if absolutely necessary and ensure redaction.

Example:
```go
// Before
logs.CtxInfo(ctx, "GenAnalysisReport get csv url=%v", url)

// After (safer)
logs.CtxInfo(ctx, "GenAnalysisReport prepared csv file=%s", fileName)
// or redact
if u, err := neturl.Parse(url); err == nil {
    logs.CtxInfo(ctx, "GenAnalysisReport csv path=%s", u.Path)
}
```

Reference: Go Code Review Comments (logging), and internal security best practices: do not log secrets or signed URLs.

backend/modules/evaluation/domain/service/insight_analysis_impl.go
```diff
@@ -0,0 +1,322 @@
...
+
+	reportID, err := e.agentAdapter.CallTraceAgent(ctx, spaceID, url)
+	if err != nil {
+		return err
+	}
+	logs.CtxInfo(ctx, "[GenAnalysisReport] CallTraceAgent success, expt_id=%v, record_id=%v, report_id=%v, csv url=%v", exptID, recordID, reportID, url)
```
CozeLoop: [Should-Fix] Redact secret-bearing data in logs

This log line includes the same pre-signed URL. For the same reasons as above, please avoid logging the full URL (tokens/credentials in query string). Prefer logging only fileName or a redacted path:

```go
// Before
logs.CtxInfo(ctx, "[GenAnalysisReport] CallTraceAgent success, expt_id=%v, record_id=%v, report_id=%v, csv url=%v", exptID, recordID, reportID, url)

// After
logs.CtxInfo(ctx, "[GenAnalysisReport] CallTraceAgent success, expt_id=%v, record_id=%v, report_id=%v, csv_file=%s", exptID, recordID, reportID, fileName)
```

This aligns with secure logging practices and reduces leakage surface.

backend/modules/evaluation/domain/service/expt_result_impl.go
```diff
@@ -619,6 +619,7 @@ func (e ExptResultServiceImpl) getColumnAnnotations(ctx context.Context, spaceID
 		}
 		exptColumnAnnotations = append(exptColumnAnnotations, exptColumnAnnotation)
 	}
+	ctx = contexts.WithCtxWriteDB(ctx)
```
CozeLoop: [Should-Fix] Unconditionally switching to write DB for a read path may cause unnecessary load and latency

Context: getColumnAnnotations sets the context to write-DB via contexts.WithCtxWriteDB(ctx) before BatchGetExptTurnResultTagRefs. This method appears to be a pure read. Forcing reads to the writer can overload the primary and increase latency.

Recommendation: Only opt into write-DB routing when you have a clear read-after-write consistency requirement signaled by the caller (e.g., a flag or lwt/check). Otherwise, use the default read pool.

Example options:
- Remove the WithCtxWriteDB call here and rely on the existing latest-write tracker used elsewhere (e.lwt.CheckWriteFlagByID) to decide when to route to writer.
- Or gate it behind a condition:
```go
if needStrongConsistency { // e.g., detected from param/lwt/ctx
    ctx = contexts.WithCtxWriteDB(ctx)
}
```

This aligns with performance best practices and avoids turning benign read paths into writer-heavy operations.

idl/thrift/coze/loop/evaluation/coze.loop.evaluation.expt.thrift
```diff
@@ -451,6 +451,101 @@ struct GetExptResultExportRecordResponse {
...
+struct DeleteExptInsightAnalysisRecordResponse {
+
+    255: base.BaseResp BaseResp
+}
+
+struct FeedbackExptInsightAnalysisReportRequest {
+    1: required i64 workspace_id (api.body = 'workspace_id', api.js_conv = 'true', go.tag = 'json:"workspace_id"')
+    2: required i64 expt_id (api.path = 'expt_id' , api.js_conv = 'true', go.tag = 'json:"expt_id"')
+    3: required i64 insight_analysis_record_id (api.path = 'insight_analysis_record_id', api.js_conv = 'true', go.tag = 'json:"insight_analysis_record_id"')
+    4: required expt.FeedbackActionType FeedbackActionType
```
CozeLoop: [Must-Fix] Thrift field name must use snake_case per IDL spec

Problem: The field `FeedbackActionType` uses PascalCase. Per CozeLoop IDL specification (字段命名使用蛇形命名法), all struct fields must be snake_case. Keeping mixed casing will break consistency and codegen conventions across SDKs.

Recommendation: Rename the field to `feedback_action_type` and keep the type as `expt.FeedbackActionType`. Update server/client mappings accordingly (converter and handler code already treat this as a separate value, so renaming the field should be localized to IDL and generated code use sites).

Example:
```thrift
struct FeedbackExptInsightAnalysisReportRequest {
    1: required i64 workspace_id
    2: required i64 expt_id (api.path = 'expt_id', ...)
    3: required i64 insight_analysis_record_id (api.path = 'insight_analysis_record_id', ...)
    4: required expt.FeedbackActionType feedback_action_type
    5: optional string comment
    6: optional i64 comment_id (api.body = 'comment_id', ...)
    200: optional common.Session session
    255: optional base.Base Base
}
```

Reference: CozeLoop_IDL规范.lark.md §4 结构体字段命名。

release/deployment/docker-compose/bootstrap/mysql-init/init-sql/expt_insight_analysis_record.sql
```diff
@@ -0,0 +1,14 @@
+CREATE TABLE IF NOT EXISTS `expt_insight_analysis_record` (
```
CozeLoop, line 1: [Must-Fix] Helm chart init SQL parity missing for new tables

Problem: This PR adds new tables under docker-compose path (expt_insight_analysis_record/feedback_vote/feedback_comment), but the corresponding Helm chart init-sql directory does not include the same DDLs. This causes inconsistent deployments: Kubernetes installs will miss these tables.

Action: Add equivalent SQL files to:
`release/deployment/helm-chart/charts/app/bootstrap/init/mysql/init-sql/`
so both docker-compose and Helm paths are aligned (per DB change guidelines). Ensure idempotency and indexes match.

Tables to add:
- expt_insight_analysis_record.sql
- expt_insight_analysis_feedback_vote.sql
- expt_insight_analysis_feedback_comment.sql

Reference: 数据库规范 — 保持 docker/k8s 初始化脚本一致。

backend/modules/evaluation/infra/repo/experiment/mysql/expt_insight_analysis_feedback_comment.go
```diff
@@ -0,0 +1,98 @@
...
+
+	return nil
+}
+
+func (e exptInsightAnalysisFeedbackCommentDAO) List(ctx context.Context, spaceID, exptID, recordID int64, page entity.Page) ([]*model.ExptInsightAnalysisFeedbackComment, int64, error) {
+	var (
+		finds []*model.ExptInsightAnalysisFeedbackComment
+		total int64
+	)
+	db := e.db.NewSession(ctx).Model(&model.ExptInsightAnalysisFeedbackComment{}).
```
CozeLoop, line 85: [Suggestion] Add explicit ordering for stable pagination

Currently the List query does not specify ORDER BY. Pagination with OFFSET/LIMIT on an unordered set can yield non-deterministic page contents across calls (especially under concurrent writes).

Recommendation: Add an index-aligned ORDER BY (e.g., created_at DESC or id DESC) before Count/Limit to ensure stable, predictable ordering:

```go
db := e.db.NewSession(ctx).Model().Where(...)
    .Order("created_at DESC")
// Count/Limit follows
```

This improves UX and avoids flakiness in clients consuming comment lists.

CozeLoop (changes requested):
Summary: Thank you for the substantial feature work on Insight Analysis. I’ve focused the review on correctness, security, and operational parity. I filed several concrete issues with code-level suggestions:

Must-Fix
- Prevent persisting AnalysisReportID when it’s zero to avoid invalid polling
- IDL field naming: use snake_case (feedback_action_type) per spec
- Kubernetes Helm SQL parity for the three new tables

Should-Fix
- Avoid logging pre-signed URLs (two locations)
- Don’t unconditionally route read paths to write-DB; gate on consistency needs

Suggestion
- Add ORDER BY to feedback comment listing for stable pagination

Once these are addressed, the change looks solid to move forward. Thanks!

backend/modules/evaluation/domain/service/insight_analysis_impl.go
CozeLoop: [Must-Fix] Persisting a non-nil AnalysisReportID when reportID is zero creates an invalid polling state

Problem: In the defer block, AnalysisReportID is always set to a non-nil pointer, even when reportID is still 0 due to an early error. On the next scheduled check, checkAnalysisReportGenStatus will treat it as an existing report and call GetReport(ctx, spaceID, 0), which is invalid and may cause noisy logs or erroneous state transitions.

Why it matters: This introduces incorrect behavior under failure paths and makes the status machine harder to reason about. Effective Go and Go Code Review Comments recommend clear, consistent error handling and avoiding sentinel values that look valid.

Fix: Only set AnalysisReportID when reportID > 0. Example:

```go
        analysisReportID := int64(0)
        defer func() {
            record := &entity.ExptInsightAnalysisRecord{
                ID:                 recordID,
                SpaceID:            spaceID,
                ExptID:             exptID,
                ExptResultFilePath: ptr.Of(exptResultFilePath),
                Status:             entity.InsightAnalysisStatus_Running,
            }
            if analysisReportID > 0 {
                record.AnalysisReportID = ptr.Of(analysisReportID)
            }
            if err != nil {
                record.Status = entity.InsightAnalysisStatus_Failed
            }
            if err1 := e.repo.UpdateAnalysisRecord(ctx, record); err1 != nil {
                logs.CtxError(ctx, "UpdateAnalysisRecord failed: %v", err1)
                err = err1
            }
        }()
```

backend/modules/evaluation/domain/service/insight_analysis_impl.go
CozeLoop: [Should-Fix] Pre-signed URL logged in plaintext (leaks secret-bearing query strings)

Problem: The pre-signed CSV download URL is logged at info level. These URLs typically contain time-limited credentials or signatures in the query string. Emitting them to logs can leak access to private data and violates secure logging practices.

Guidance: Avoid logging full pre-signed URLs. If logging is necessary for debugging, either omit the query string or redact sensitive parts.

Fix: Mask the query parameters before logging, or log only the path/filename. Example:

```go
// before
logs.CtxInfo(ctx, "GenAnalysisReport get csv url=%v", url)

// after (redact query string)
if u, perr := neturl.Parse(url); perr == nil {
    u.RawQuery = "<redacted>"
    logs.CtxInfo(ctx, "GenAnalysisReport get csv url=%s", u.String())
} else {
    logs.CtxInfo(ctx, "GenAnalysisReport get csv url (parse-error): %v", perr)
}
```

References: Go Code Review Comments (logging), and internal security best practices.

backend/modules/evaluation/domain/service/expt_result_impl.go
CozeLoop: [Should-Fix] Unconditionally forcing writes to primary for a read path increases load and latency

Problem: getColumnAnnotations unconditionally switches to the write DB via contexts.WithCtxWriteDB(ctx). This read operation will always hit the primary, increasing load and tail latency unnecessarily.

Guidance: Only route to primary when strict read-after-write consistency is required and the data cannot be served from replicas. Otherwise, leave the default read path.

Fix: Gate the write-DB context behind a well-defined condition (e.g., a just-updated tag reference or an explicit strong-consistency flag). Example:

```go
// before
ctx = contexts.WithCtxWriteDB(ctx)

// after
if needStrongConsistency { // e.g., recent writes in the same request
    ctx = contexts.WithCtxWriteDB(ctx)
}
```

This aligns with CozeLoop server code guidelines on efficiency and resource usage, and with Go Code Review Comments on avoiding unnecessary work.

idl/thrift/coze/loop/evaluation/coze.loop.evaluation.expt.thrift
CozeLoop: [Must-Fix] Thrift field names must be snake_case per IDL spec; PascalCase breaks conventions and generators

Problem: Field name "FeedbackActionType" uses PascalCase. CozeLoop IDL规范 requires snake_case for field identifiers to ensure consistent code generation and API contracts across languages.

Fix: Rename to snake_case, e.g. `feedback_action_type`.

```thrift
struct FeedbackExptInsightAnalysisReportRequest {
    1: required i64 workspace_id
    2: required i64 expt_id
    3: required i64 insight_analysis_record_id
    4: required expt.FeedbackActionType feedback_action_type // rename
    5: optional string comment
    6: optional i64 comment_id
}
```

Reference: CozeLoop_IDL规范（命名与风格章节）。

release/deployment/docker-compose/bootstrap/mysql-init/init-sql/expt_insight_analysis_record.sql
CozeLoop, line 1: [Must-Fix] Init SQL added for Docker Compose but missing in Helm chart init scripts (deployment parity)

Problem: This new table DDL exists under docker-compose bootstrap, but no corresponding DDL is added under Helm at `release/deployment/helm-chart/charts/app/bootstrap/init/mysql`. This causes environment drift: Kubernetes deployments will miss required tables.

Fix: Add the same DDL to the Helm init SQL directory, aligning with our database change guidelines. Ensure both Docker Compose and Helm paths are updated together.

Reference: 数据库规范（多部署方式 SQL 变更一致性要求）。

release/deployment/docker-compose/bootstrap/mysql-init/init-sql/expt_insight_analysis_feedback_vote.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE IF NOT EXISTS `expt_insight_analysis_feedback_vote` (
```
CozeLoop, line 1: [Must-Fix] Init SQL added for Docker Compose but missing in Helm chart init scripts (deployment parity)

Problem: This new table DDL exists under docker-compose bootstrap, but no corresponding DDL is added under Helm at `release/deployment/helm-chart/charts/app/bootstrap/init/mysql`. This causes environment drift: Kubernetes deployments will miss required tables.

Fix: Add the same DDL to the Helm init SQL directory, aligning with our database change guidelines. Ensure both Docker Compose and Helm paths are updated together.

Reference: 数据库规范（多部署方式 SQL 变更一致性要求）。

release/deployment/docker-compose/bootstrap/mysql-init/init-sql/expt_insight_analysis_feedback_comment.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE IF NOT EXISTS `expt_insight_analysis_feedback_comment` (
```
CozeLoop, line 1: [Must-Fix] Init SQL added for Docker Compose but missing in Helm chart init scripts (deployment parity)

Problem: This new table DDL exists under docker-compose bootstrap, but no corresponding DDL is added under Helm at `release/deployment/helm-chart/charts/app/bootstrap/init/mysql`. This causes environment drift: Kubernetes deployments will miss required tables.

Fix: Add the same DDL to the Helm init SQL directory, aligning with our database change guidelines. Ensure both Docker Compose and Helm paths are updated together.

Reference: 数据库规范（多部署方式 SQL 变更一致性要求）。

backend/modules/evaluation/infra/repo/experiment/mysql/expt_insight_analysis_feedback_comment.go
```diff
@@ -0,0 +1,98 @@
...
+		Where("space_id =?", spaceID).
+		Where("expt_id =?", exptID).
+		Where("analysis_record_id =?", recordID)
+	// 总记录数
+	db = db.Count(&total)
+	// 分页
+	db = db.Offset(page.Offset()).Limit(page.Limit())
+	err := db.Find(&finds).Error
```
CozeLoop, line 93: [Suggestion] Add an explicit ORDER BY for stable pagination of comments

Problem: The List query paginates without a deterministic order. Without ORDER BY, result ordering may be non-deterministic across pages and DB engines, leading to duplicate/missed rows under pagination.

Fix: Add an explicit order (e.g., created_at DESC or id DESC) before Offset/Limit.

```go
db := e.db.NewSession(ctx).Model([0;49;39m[0m[0;49;39m[0m[0m[0m[m[0m[0;49;39m[0m[0m[0m[0;49;39m[0m[0m[0m[0m[0m[0;49;39m[0m[0m[0m[0;49;39m[0m[0m[0m[0m[0;49;39m[0m[0m[0m[0m[0;49;39m[0m[0m[0m[0m[0m[0;49;39m[0m[0m[0m[0;49;39m[0m[0m[0m[0m[0;49;39m[0m[0m[0m[0m[0;49;39m[0m[0m[0m[0m[0m[0;49;39m[0m[0m[0m[0;49;39m[0m[0m[0m[0m[0m[0;49;39m[0m[0m[0m[0;49;39m[0m[0m[0m[0m[0;49;39m[0m[0m[0m[0m[m[0m[0;49;39m[0m[0m[0m[0m[m[0m[0;49;39m[0m[0m[0m[0m[m[0m[0;49;39m[0m[0m[0m[0m
    // ... where conditions ...
).Order("created_at DESC")
```

This aligns with Go Code Review Comments on determinism and pagination best practices.

backend/modules/evaluation/infra/rpc/agent/agent.go
```diff
@@ -0,0 +1,26 @@
...
+	return 0, errors.New("CallTraceAgent not implement")
+}
+
+func (a AgentAdapter) GetReport(ctx context.Context, spaceID, reportID int64) (report string, status entity.ReportStatus, err error) {
+	return "", 0, errors.New("CallTraceAgent not implement")
```
CozeLoop: [Should-Fix] Stub implementation returns generic errors; consider clear error typing and message, or feature flag

Problem: Both CallTraceAgent and GetReport return a generic error (and GetReport’s error message says "CallTraceAgent not implement"). Shipping this adapter via wire binding means the new feature will always fail at runtime.

Fix options:
- Implement the adapter before exposing APIs; or
- Guard the feature behind a config flag so endpoints aren’t callable until ready; and
- At minimum, fix the error message and convert to a typed/structured errno to avoid leaking implementation detail.

Example:
```go
return "", 0, errno.NewFeatureUnavailable("GetReport not implemented")
```

CozeLoop (changes requested):
Summary: Insight Analysis feature is a solid direction and wires most components end-to-end. I found several correctness and consistency issues that should be addressed before merge.

Must-Fix:
1) Avoid persisting AnalysisReportID when reportID == 0 (invalid polling state). See inline.
2) Thrift field name should be snake_case (feedback_action_type) per IDL规范. See inline.
3) Add Helm init-sql parity for the three new insight tables to avoid K8s drift. See inline.

Should-Fix:
1) Don’t log full pre-signed URLs; redact query strings. See inline.
2) Don’t unconditionally route read paths to write DB for column annotations. See inline.
3) RPC adapter is a stub; either implement, guard with feature flag, or return a typed errno, and fix GetReport error message.

Suggestions:
1) Add ORDER BY to feedback comments listing for stable pagination.

Once these are addressed, I’m happy to re-review quickly.

CozeLoop (commented):
Verification: actions executed on PR #191 — inline comments and CHANGES_REQUESTED posted

Hi,

This is a quick proof-of-action note to confirm that a full code review has been executed and submitted on this PR.

What was done:
- Submitted two reviews with status CHANGES_REQUESTED (latest at 2025-09-18T12:07:27Z, commit cb47d50bc9c8aa98589ffb22e1a042bd0308eb57)
- Posted multiple inline comments on changed lines (Side=RIGHT), including (non-exhaustive):
  - backend/modules/evaluation/domain/service/insight_analysis_impl.go:97 — [Must-Fix] avoid persisting AnalysisReportID when reportID==0
  - idl/thrift/coze/loop/evaluation/coze.loop.evaluation.expt.thrift:519 — [Must-Fix] field name must be snake_case
  - release/deployment/docker-compose/bootstrap/mysql-init/init-sql/expt_insight_analysis_record.sql:1 — [Must-Fix] add Helm init-sql parity
  - backend/modules/evaluation/domain/service/expt_result_impl.go:622 — [Should-Fix] don’t force read path to primary
  - backend/modules/evaluation/infra/rpc/agent/agent.go:25 — [Should-Fix] stub impl / error message
  - backend/modules/evaluation/infra/repo/experiment/mysql/expt_insight_analysis_feedback_comment.go:86-93 — [Suggestion] add ORDER BY for stable pagination

How to view:
- Conversation tab: see timeline entries “CozeLoop requested changes” (most recent 12:07:27Z)
- Files changed tab: ensure diff shows all changes (no outdated-only filter); inline comments appear on the RIGHT side of modified lines

Full CR report:
- Feishu document (internal): https://bytedance.larkoffice.com/docx/AhamdrpGsoEmtKxtO3icqqKanzb

If anything still doesn’t show on your side, please refresh the PR page and ensure you’re viewing the latest head (cb47d50…). Happy to re-review quickly after fixes are pushed.

backend/modules/evaluation/domain/service/insight_analysis_impl.go
```diff
@@ -0,0 +1,323 @@
...
+		return e.repo.UpdateFeedbackVote(ctx, feedbackVote)
+	case entity.FeedbackActionType_Downvote:
+		feedbackVote := &entity.ExptInsightAnalysisFeedbackVote{
+			SpaceID:          param.SpaceID,
+			ExptID:           param.ExptID,
+			AnalysisRecordID: param.AnalysisRecordID,
+			CreatedBy:        param.Session.UserID,
+			VoteType:         entity.Downvote,
+		}
+		return e.repo.CreateFeedbackVote(ctx, feedbackVote)
```
CozeLoop, line 294: [🚫Must-Fix] Feedback vote creation will violate unique constraint and fail on repeated actions (lines 268-294)

Problem: Both Upvote and Downvote branches always call CreateFeedbackVote, which inserts a new row regardless of whether a vote by the same user already exists. The table expt_insight_analysis_feedback_vote has a UNIQUE constraint (space_id, expt_id, analysis_record_id, created_by). If a user has already voted (or even canceled), subsequent Upvote/Downvote will try to insert again and will fail with a duplicate key error instead of updating the existing row.

Why this matters: Correctness and UX. Users won’t be able to toggle their vote reliably; retries will fail and the API will return errors instead of switching the vote. This also risks partial failures in concurrent scenarios.

Fix suggestion: Upsert vote instead of unconditional insert. For example: first attempt UpdateFeedbackVote; if no rows affected, then CreateFeedbackVote; or use an atomic upsert at the DAO layer (GORM OnConflict) keyed by (space_id, expt_id, analysis_record_id, created_by).

Example (service-level logic):
```go
existing, err := e.repo.GetFeedbackVoteByUser(ctx, param.SpaceID, param.ExptID, param.AnalysisRecordID, param.Session.UserID)
if err != nil { return err }
feedbackVote := &entity.ExptInsightAnalysisFeedbackVote{
    SpaceID: param.SpaceID, ExptID: param.ExptID,
    AnalysisRecordID: param.AnalysisRecordID,
    CreatedBy: param.Session.UserID, VoteType: entity.Upvote, // or Downvote
}
if existing != nil {
    return e.repo.UpdateFeedbackVote(ctx, feedbackVote)
}
return e.repo.CreateFeedbackVote(ctx, feedbackVote)
```
Alternatively (DAO):
```go
// Using GORM clause.OnConflict to do: INSERT ... ON DUPLICATE KEY UPDATE vote_type=excluded.vote_type
```

backend/modules/evaluation/domain/service/insight_analysis_impl.go
```diff
@@ -0,0 +1,323 @@
...
+			ID:               ptr.From(param.CommentID),
+			SpaceID:          param.SpaceID,
+			ExptID:           param.ExptID,
+			AnalysisRecordID: param.AnalysisRecordID,
+			CreatedBy:        param.Session.UserID,
+			Comment:          ptr.From(param.Comment),
+		}
+		return e.repo.UpdateFeedbackComment(ctx, feedbackComment)
+	case entity.FeedbackActionType_Delete_Comment:
+		return e.repo.DeleteFeedbackComment(ctx, param.SpaceID, param.ExptID, ptr.From(param.CommentID))
```
CozeLoop, line 321: [🚫Must-Fix] Missing validation for CommentID leads to unintended updates/deletes (lines 304-315)

Problem: Update_Comment and Delete_Comment paths call ptr.From(param.CommentID) without checking if CommentID is nil. If the client forgets to pass CommentID, ptr.From(nil) returns the zero value (0), and the DAO will attempt to update/delete comment id=0, which is incorrect and may affect nothing silently or cause unexpected behavior.

Why this matters: Correctness and safety. This can result in silent no-ops, confusing client behavior, or potential accidental updates if id=0 could ever exist in certain environments.

Fix suggestion: Validate inputs and return a clear error when CommentID is required but missing.

Example:
```go
case entity.FeedbackActionType_Update_Comment:
    if param.CommentID == nil || param.Comment == nil || *param.Comment == "" {
        return errorx.NewByCode(errno.CommonInvalidParamCode, errorx.WithExtraMsg("empty comment_id or comment"))
    }
    // proceed...
case entity.FeedbackActionType_Delete_Comment:
    if param.CommentID == nil {
        return errorx.NewByCode(errno.CommonInvalidParamCode, errorx.WithExtraMsg("empty comment_id"))
    }
    return e.repo.DeleteFeedbackComment(ctx, param.SpaceID, param.ExptID, ptr.From(param.CommentID))
```

backend/modules/evaluation/application/experiment_app.go
```diff
@@ -1077,3 +1080,202 @@ func (e *experimentApplication) GetExptResultExportRecord(ctx context.Context, r
...
+		return nil, err
+	}
+
+	err = e.auth.AuthorizationWithoutSPI(ctx, &rpc.AuthorizationWithoutSPIParam{
+		ObjectID:        strconv.FormatInt(req.GetExptID(), 10),
+		SpaceID:         req.GetWorkspaceID(),
+		ActionObjects:   []*rpc.ActionObject{{Action: gptr.Of(consts.Edit), EntityType: gptr.Of(rpc.AuthEntityType_EvaluationExperiment)}},
+		OwnerID:         gptr.Of(got.CreatedBy),
+		ResourceSpaceID: req.GetWorkspaceID(),
+	})
```
CozeLoop: [🚨Should-Fix] Listing records requires Edit permission; consider using Read to avoid over-restricting access (lines 1133-1139)

Problem: ListExptInsightAnalysisRecord checks Edit permission on the experiment. Listing (read-only) data typically should use read-level auth (e.g., ActionReadExpt on space or experiment) so that viewers can see analysis results without edit rights.

Why this matters: Maintainability and authorization semantics. Overusing Edit for read paths prevents legitimate read-only users from accessing data and causes unnecessary permission escalations.

Fix suggestion: Replace the Edit action with a read action consistent with the existing policy used in GetExptInsightAnalysisRecord/ListExptInsightAnalysisComment (they use ActionReadExpt). For example:
```go
err = e.auth.Authorization(ctx, 6rpc.AuthorizationParam{
    ObjectID: strconv.FormatInt(req.WorkspaceID, 10),
    SpaceID:  req.WorkspaceID,
    ActionObjects: []*rpc.ActionObject{{Action: gptr.Of(consts.ActionReadExpt), EntityType: gptr.Of(rpc.AuthEntityType_Space)}},
})
```
If per-experiment read is required, provide a corresponding read action at experiment scope rather than Edit.

backend/modules/evaluation/domain/service/insight_analysis_impl.go
```diff
@@ -0,0 +1,323 @@
...
+	if len(userInfos) != 1 || userInfos[0] == nil {
+		return nil
+	}
+
+	userInfo := userInfos[0]
+	err = e.notifyRPCAdapter.SendMessageCard(ctx, ptr.From(userInfo.Email), consts.InsightAnalysisNotifyCardID, map[string]string{
+		"expt_name": expt.Name,
+		"space_id":  strconv.FormatInt(spaceID, 10),
+		"expt_id":   strconv.FormatInt(exptID, 10),
+	})
```
CozeLoop, line 250: [💡Suggestion] Consider using a stable user identifier instead of email for notifications (lines 245-250)

Observation: SendMessageCard is called with userInfo.Email. If the notify service expects a user ID (or UID) rather than an email, this may silently fail or deliver to the wrong account in organizations where email != principal ID.

Suggestion: Pass the canonical user identifier that your notify system expects (e.g., userInfo.UserID or UID) and include email as a fallback only if the adapter is email-based.

Example:
```go
recipient := userInfo.UserID // or the ID field your notify system requires
if recipient == "" {
    recipient = userInfo.Email
}
return e.notifyRPCAdapter.SendMessageCard(ctx, recipient, consts.InsightAnalysisNotifyCardID, map[string]string{ ... })
```

HymanShi (approved):

PR #253, opened 2025-10-23
[fix][infra]modify special benefit struct

xueyizheng:
#### What type of PR is this?

feat: change a benefit struct to fit special task
#### Check the PR title.

en:
zh(optional):

### DIFF
--- a/backend/infra/external/benefit/benifit.go
+++ b/backend/infra/external/benefit/benifit.go
@@ -28,7 +28,7 @@ type IBenefitService interface {
 	BatchCheckEnableTypeBenefit(ctx context.Context, param *BatchCheckEnableTypeBenefitParams) (result *BatchCheckEnableTypeBenefitResult, err error)
 	// CheckAndDeductOptimizationBenefit 校验扣减优化权益
 	CheckAndDeductOptimizationBenefit(ctx context.Context, param *CheckAndDeductOptimizationBenefitParams) (result *CheckAndDeductOptimizationBenefitResult, err error)
-	// DeductOptimizationBenefit 上报优化资源点
+	// Deprecated: DeductOptimizationBenefit is deprecated. Use CheckAndDeductOptimizationBenefit(...) instead.
 	DeductOptimizationBenefit(ctx context.Context, param *DeductOptimizationBenefitParams) (err error)
 }
 
@@ -111,6 +111,14 @@ func (h *DenyReason) ToErr() error {
 	}
 }
 
+type When int64
+
+const (
+	WhenStart   When = 1
+	WhenRunning When = 2
+	WhenFinish  When = 3
+)
+
 type CheckAndDeductEvalBenefitParams struct {
 	ConnectorUID string            `json:"connector_uid"` // Coze登录ID
 	SpaceID      int64             `json:"space_id"`      // 空间ID
@@ -146,10 +154,12 @@ type BatchCheckEnableTypeBenefitResult struct {
 }
 
 type CheckAndDeductOptimizationBenefitParams struct {
-	ConnectorUID string `json:"connector_uid"` // Coze登录ID
-	SpaceID      int64  `json:"space_id"`      // 空间ID
-	PromptID     int64  `json:"prompt_id"`     // prompt id，用于唯一标识
-	TaskID       int64  `json:"task_id"`       // task id
+	ConnectorUID string  `json:"connector_uid"` // Coze登录ID
+	SpaceID      int64   `json:"space_id"`      // 空间ID
+	PromptID     int64   `json:"prompt_id"`     // prompt id，用于唯一标识
+	TaskID       int64   `json:"task_id"`       // task id
+	Amount       float64 `json:"amount"`        // 消耗的资源点数
+	When         When    `json:"when"`          // 适用场景：1-启动时校验，2-运行时校验，3-结束时校验
 }
 
 type CheckAndDeductOptimizationBenefitResult struct {
--- a/backend/infra/platestwrite/latest_write_tracker.go
+++ b/backend/infra/platestwrite/latest_write_tracker.go
@@ -129,6 +129,7 @@ const (
 	ResourceTypePromptCommit             ResourceType = "prompt_commit"
 	ResourceTypePromptLabel              ResourceType = "prompt_label"
 	ResourceTypePromptCommitLabelMapping ResourceType = "prompt_commit_label_mapping"
+	ResourceTypeCozeloopOptimizeTask     ResourceType = "cozeloop_optimize_task" // 外场智能优化
 
 	ResourceTypeExperiment    ResourceType = "experiment"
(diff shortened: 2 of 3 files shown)


backend/infra/external/benefit/benifit.go
```diff
@@ -111,6 +111,14 @@ func (h *DenyReason) ToErr() error {
 	}
 }
 
+type When int64
```
wanglizhi, line 114: 这个是用来做什么的？枚举不是都用String吗

backend/infra/external/benefit/benifit.go
```diff
@@ -146,10 +154,12 @@ type BatchCheckEnableTypeBenefitResult struct {
...
 type CheckAndDeductOptimizationBenefitParams struct {
-	ConnectorUID string `json:"connector_uid"` // Coze登录ID
-	SpaceID      int64  `json:"space_id"`      // 空间ID
-	PromptID     int64  `json:"prompt_id"`     // prompt id，用于唯一标识
-	TaskID       int64  `json:"task_id"`       // task id
+	ConnectorUID string  `json:"connector_uid"` // Coze登录ID
+	SpaceID      int64   `json:"space_id"`      // 空间ID
+	PromptID     int64   `json:"prompt_id"`     // prompt id，用于唯一标识
+	TaskID       int64   `json:"task_id"`       // task id
+	Amount       float64 `json:"amount"`        // 消耗的资源点数
```
CozeLoop, line 161: **[🚨Should-Fix] Inconsistent type for resource points**

Problem: This params struct uses `float64` for `Amount`, while `DeductOptimizationBenefitParams` uses `int64` (`ResourcePoints`). Representing "resource points" as floating-point can introduce precision/rounding errors and make units inconsistent across APIs.

Why it matters: Precision mismatches will cause subtle billing/accounting bugs, and complicate downstream validations. ByteDance Go specs recommend using integer base units for currency/points to avoid FP precision issues.

Proposal: Align on integer base units. Either:
- Change `Amount` to `int64` and rename to `ResourcePoints` for consistency; or
- If fractional points are required, store fixed-point as `int64` (e.g., milli-points) and convert at boundaries.

Example:
```go
// Prefer integer base units to avoid FP precision
type CheckAndDeductOptimizationBenefitParams struct {
    ConnectorUID   string `json:"connector_uid"`
    SpaceID        int64  `json:"space_id"`
    PromptID       int64  `json:"prompt_id"`
    TaskID         int64  `json:"task_id"`
-   Amount         float64 `json:"amount"`
+   ResourcePoints int64   `json:"resource_points"`
    When           When    `json:"when"`
}
```

backend/infra/external/benefit/benifit.go
```diff
@@ -146,10 +154,12 @@ type BatchCheckEnableTypeBenefitResult struct {
...
+	When         When    `json:"when"`          // 适用场景：1-启动时校验，2-运行时校验，3-结束时校验
```
CozeLoop, line 162: **[💡Suggestion] Clarify stage enum and zero-value semantics for `When`**

Problem: The `When` field relies on a custom `When` enum with values starting at 1. The zero value of the type is currently invalid/unspecified, which can lead to ambiguous behavior when the field is omitted or not initialized.

Why it matters: Clear zero-value semantics improve robustness of JSON decoding and defaulting. Small, well-scoped enums using `iota` are idiomatic in Go and are easier to maintain.

Proposal:
- Introduce an explicit `WhenUnknown = 0` to represent "unspecified", and document the valid range.
- Prefer `int` with `iota` unless 64-bit width is strictly required (keeps types lightweight).
- Optionally rename `When` to a more descriptive name like `OptimizationStage` to improve readability.

Example:
```go
// Use iota and explicit zero value
type OptimizationStage int
const (
    WhenUnknown OptimizationStage = iota
    WhenStart
    WhenRunning
    WhenFinish
)

// In params
type CheckAndDeductOptimizationBenefitParams struct {
    // ...
    When OptimizationStage `json:"when"` // 0: unknown (omitted), 1: start, 2: running, 3: finish
}
```

backend/infra/external/benefit/benifit.go
```diff
@@ -111,6 +111,14 @@ func (h *DenyReason) ToErr() error {
 	}
 }
 
+type When int64
+
+const (
+	WhenStart   When = 1
+	WhenRunning When = 2
+	WhenFinish  When = 3
+)
```
CozeLoop, line 120: **[💡Suggestion] Use `iota` for enum and provide descriptive type name**

Problem: `type When int64` with manually assigned constants is valid, but it obscures default/zero-value behavior and uses 64-bit width without necessity.

Why it matters: Idiomatic Go enums use `iota` (compact, clearer), and explicit zero values prevent accidental misuse. A descriptive type (e.g., `OptimizationStage`) reduces cognitive load.

Proposal:
```go
// Descriptive enum with iota and zero-value defined
type OptimizationStage int
const (
    WhenUnknown OptimizationStage = iota // zero value: unspecified
    WhenStart                            // 1
    WhenRunning                          // 2
    WhenFinish                           // 3
)
```
- Switch the param field to `OptimizationStage`.
- Document meaning of each stage in the comment.

backend/infra/platestwrite/latest_write_tracker.go
```diff
@@ -129,6 +129,7 @@ const (
 	ResourceTypePromptCommit             ResourceType = "prompt_commit"
 	ResourceTypePromptLabel              ResourceType = "prompt_label"
 	ResourceTypePromptCommitLabelMapping ResourceType = "prompt_commit_label_mapping"
+	ResourceTypeCozeloopOptimizeTask     ResourceType = "cozeloop_optimize_task" // 外场智能优化
```
CozeLoop, line 132: **[💡Suggestion] Confirm downstream usage and naming consistency for new resource type**

Observation: Added `ResourceTypeCozeloopOptimizeTask = "cozeloop_optimize_task"` for latest-write tracking.

Why it matters: Resource types feed into Redis key generation (`db_last_write_flag:*`), so naming changes affect cross-service conventions and observability. Ensuring consistent usage across readers/writers prevents cache misses.

Proposal:
- Verify that all write paths for optimize tasks call `SetWriteFlag(...)` with this `ResourceType` (both ID-based and search-based forms).
- Double-check consumers (read paths) use the same string when calling `CheckWriteFlag...` to avoid key mismatches.
- Consider aligning name with existing patterns (e.g., `optimize_task` vs `optimization_task`) if there's a documented convention.

backend/infra/external/benefit/benifit.go
xueyizheng: 这块是chunxu定义的，我理解与fornax_infra_sdk那边有关系

backend/infra/external/benefit/benifit.go
xueyizheng: 实际使用就是float，旧的不对

PR #259, opened 2025-10-27
[fix][backend]: search trace tree idl

cuichen01:
#### Check the PR title.

en:
zh(optional):

### DIFF
--- a/release/deployment/docker-compose/conf/observability.yaml
+++ b/release/deployment/docker-compose/conf/observability.yaml
@@ -329,25 +329,17 @@ query_trace_rate_limit_config:
   space_max_qps:
     123456: 100
 
-key_columns:
-  - "start_time"
-  - "span_id"
-  - "parent_id"
-  - "duration"
-  - "span_type"
-  - "span_name"
-  - "status_code"
-  - "tags_long"
-  - "logic_delete_date"
-
 key_span_type:
   default:
-    - "model",
-    - "tool",
+    - "model"
+    - "tool"
     - "root"
     - "UserInput"
     - "UserInputV2"
     - "LLMCall"
+    - "WorkflowLLMCall"
+    - "WorkflowPluginTool"
+    - "PluginTool"
 
 correction_mq_consumer_config:
   addr:
--- a/release/deployment/helm-chart/umbrella/conf/observability.yaml
+++ b/release/deployment/helm-chart/umbrella/conf/observability.yaml
@@ -342,12 +342,15 @@ key_columns:
 
 key_span_type:
   default:
-    - "model",
-    - "tool",
+    - "model"
+    - "tool"
     - "root"
     - "UserInput"
     - "UserInputV2"
     - "LLMCall"
+    - "WorkflowLLMCall"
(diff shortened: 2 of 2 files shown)


idl/thrift/coze/loop/observability/coze.loop.observability.openapi.thrift
```diff
@@ -74,7 +74,7 @@ struct SearchTraceOApiRequest {
     5: required i64 end_time (api.js_conv='true', go.tag='json:"end_time"', api.body="end_time") // ms
     6: required i32 limit (api.body="limit")
     8: optional common.PlatformType platform_type (api.body="platform_type")
-    9: optional list<string> span_ids (api.query="span_ids")
+    9: optional list<string> span_ids (api.body="span_ids")
```
CozeLoop, line 77: [🚨Should-Fix] Potential breaking change by moving span_ids from query to body
- File: idl/thrift/coze/loop/observability/coze.loop.observability.openapi.thrift:77
- Problem: Changing `span_ids` from `api.query="span_ids"` to `api.body="span_ids"` alters request binding and may break existing clients that pass this parameter via query. It also mixes parameter locations across versions, affecting backward compatibility and API observability.
- Suggested fix: Maintain backward compatibility for one release cycle. Options:
  1) Accept both query and body: keep the query field (deprecated) and add a new body field alias.
  2) If protocol must be body-only, document the change in release notes and add server-side fallback to read query when body is empty.
  Example (IDL option):
  ```thrift
  9: optional list<string> span_ids_query (api.query="span_ids") // deprecated
  10: optional list<string> span_ids (api.body="span_ids")
  ```
- Reference: OpenAPI/HTTP compatibility guidance — avoid breaking parameter location changes without deprecation/migration steps.

release/deployment/docker-compose/conf/observability.yaml
```diff
@@ -329,25 +329,17 @@ query_trace_rate_limit_config:
...
-  - "span_id"
-  - "parent_id"
-  - "duration"
-  - "span_type"
-  - "span_name"
-  - "status_code"
-  - "tags_long"
-  - "logic_delete_date"
-
 key_span_type:
```
CozeLoop, line 332: [💡Suggestion] Keep config parity across docker/k8s (key_columns removed only in docker-compose)
- File: release/deployment/docker-compose/conf/observability.yaml:332
- Problem: This PR removes the `key_columns` block from the docker-compose config, while it remains in the Helm config. Config drift can confuse operations and cause environment-specific behavior.
- Suggested fix: Either retain `key_columns` in docker-compose (even if unused) for parity, or explicitly deprecate it in Helm with a comment and remove in both. If unused, add a note explaining its removal to avoid confusion.
- Reference: Deployment config consistency best practices to minimize env-specific divergence.

CozeLoop (commented):
Overall: Changes are small and mostly safe (formatting in openapi.go, YAML cleanup and new span types). One API compatibility risk in IDL (span_ids moved to body) and a config parity suggestion for docker-compose vs Helm.

taoyifan89 (approved):

PR #268, opened 2025-10-28
[feat][backend] ob metrics new

kidkidkid:
#### Check the PR title.

en:
zh(optional):

### DIFF
--- a/backend/modules/observability/domain/task/service/taskexe/tracehub/span_trigger.go
+++ b/backend/modules/observability/domain/task/service/taskexe/tracehub/span_trigger.go
@@ -29,7 +29,6 @@ func (h *TraceHubServiceImpl) SpanTrigger(ctx context.Context, rawSpan *entity.R
 	span := rawSpan.RawSpanConvertToLoopSpan()
 	// 1.1 Filter out spans that do not belong to any space or bot
 	spaceIDs, botIDs, _ := h.getObjListWithTaskFromCache(ctx)
-	logs.CtxInfo(ctx, "space list: %v, bot list: %v, log_suffix=%s", spaceIDs, botIDs, logSuffix)
 	if !gslice.Contains(spaceIDs, span.WorkspaceID) && !gslice.Contains(botIDs, span.TagsString["bot_id"]) {
 		logs.CtxInfo(ctx, "no space or bot found for span, space_id=%s,bot_id=%s, log_suffix=%s", span.WorkspaceID, span.TagsString["bot_id"], logSuffix)
 		return nil
--- a/backend/modules/observability/domain/trace/repo/trace.go
+++ b/backend/modules/observability/domain/trace/repo/trace.go
@@ -69,13 +69,6 @@ type InsertAnnotationParam struct {
 	Annotations []*loop_span.Annotation
 }
 
-type UpsertAnnotationParam struct {
-	Tenant      string
-	TTL         loop_span.TTL
-	Annotations []*loop_span.Annotation
-	IsSync      bool
-}
-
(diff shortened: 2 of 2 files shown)


backend/modules/observability/domain/metric/service/metric.go
```diff
@@ -0,0 +1,547 @@
...
+					FilterFields:    req.FilterFields,
+					DrillDownFields: req.DrillDownFields,
+					StartTime:       req.StartTime,
+					EndTime:         req.EndTime,
+				}
+				resp, err := m.queryMetrics(ctx, sReq)
+				lock.Lock()
+				defer lock.Unlock()
+				if err == nil {
+					metricsResp[i] = resp
```
CozeLoop: [🚫Must-Fix] Closure captures loop variable `i` causing incorrect indexing under concurrency. In Go, loop variables are reused across iterations and captured by closures; here `metricsResp[i] = resp` may write to the wrong slot when goroutines execute after `i` changed. Use the parameter `t` passed into the outer function (or assign `ti := i` before launching) and index with that.

Suggested fix:
```
if err == nil {
    metricsResp[t] = resp
}
```
References: Go Code Review Comments – Loop variables in closures; Effective Go – "The loop variable and closures".

backend/modules/observability/infra/repo/trace.go
```diff
@@ -277,6 +290,34 @@ func (t *TraceCkRepoImpl) InsertAnnotations(ctx context.Context, param *repo.Ins
 	})
 }
 
+func (t *TraceCkRepoImpl) UpsertAnnotation(ctx context.Context, param *repo.UpsertAnnotationParam) error {
+	return nil
```
CozeLoop: [🚨Should-Fix] Unimplemented method returns nil. `UpsertAnnotation` currently returns nil without performing any upsert, which can lead to silent success and data inconsistency if the method is called.

Recommendation: either implement the upsert logic or return a clear error to prevent misuse until implemented, e.g.:
```
return errorx.NewByCode(obErrorx.CommercialCommonInternalErrorCodeCode, errorx.WithExtraMsg("UpsertAnnotation not implemented"))
```
This makes the contract explicit and avoids false positives.

backend/modules/observability/domain/task/service/taskexe/tracehub/span_trigger.go
```diff
@@ -29,6 +29,7 @@ func (h *TraceHubServiceImpl) SpanTrigger(ctx context.Context, rawSpan *entity.R
 	span := rawSpan.RawSpanConvertToLoopSpan()
 	// 1.1 Filter out spans that do not belong to any space or bot
 	spaceIDs, botIDs, _ := h.getObjListWithTaskFromCache(ctx)
+	logs.CtxInfo(ctx, "space list: %v, bot list: %v, log_suffix=%s", spaceIDs, botIDs, logSuffix)
```
CozeLoop: [💡Suggestion] Logging entire `spaceIDs` and `botIDs` slices at info level may be noisy and could expose identifiers in production logs. Consider downgrading to debug level or summarizing counts to reduce log volume:

```
logs.CtxDebug(ctx, "space_count=%d, bot_count=%d, log_suffix=%s", len(spaceIDs), len(botIDs), logSuffix)
```
This keeps observability while avoiding potentially verbose logs.

backend/modules/observability/domain/metric/service/metric.go
kidkidkid: fixed

backend/modules/observability/domain/task/service/taskexe/tracehub/span_trigger.go
kidkidkid: fixed

PR #275, opened 2025-10-31
[fix][backend] Add default trace duration config.

taoyifan89:
Change-Id: I7772559f121d08a931f1ea2d70c416d6f608abcd

#### What type of PR is this?
fix

#### Check the PR title.

en: Add a default trace duration to fix some bug.
zh(optional):

### DIFF
--- a/backend/modules/observability/infra/config/trace.go
+++ b/backend/modules/observability/infra/config/trace.go
@@ -108,10 +108,12 @@ func (t *TraceConfigCenter) GetTraceFieldMetaInfo(ctx context.Context) (*config.
 
 func (t *TraceConfigCenter) GetTraceDataMaxDurationDay(ctx context.Context, platformPtr *string) int64 {
 	defaultDuration := int64(7)
+	var platformType string
 	if platformPtr == nil {
-		return defaultDuration
+		platformType = "default"
+	} else {
+		platformType = *platformPtr
 	}
-	platformType := *platformPtr
 	mp := make(map[string]int64)
 	err := t.UnmarshalKey(ctx, traceMaxDurationDay, &mp)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


kidkidkid (approved):

backend/modules/observability/infra/config/trace.go
```diff
@@ -121,6 +121,9 @@ func (t *TraceConfigCenter) GetTraceDataMaxDurationDay(ctx context.Context, plat
 	if mp[platformType] > 0 {
 		return mp[platformType]
 	} else {
+		if duration, ok := mp["default"]; ok && duration > 0 {
+			return duration
+		}
```
CozeLoop: **[🚨Should-Fix]** Add unit test for default trace duration fallback
- File: backend/modules/observability/infra/config/trace.go:124-126
- Issue: New logic returns `mp["default"]` when a platform-specific duration is absent. This behavior change is correct but currently untested, risking silent regression if the fallback is modified later.
- Recommendation: Add a test case in `TraceConfigCenter_GetTraceDataMaxDurationDay` to assert the default fallback path. Example:
```go
mockLoader.EXPECT().UnmarshalKey(gomock.Any(), traceMaxDurationDay, gomock.Any()).
    DoAndReturn(func(ctx context.Context, key string, v interface{}, _ ...interface{}) error {
        mp := v.(*map[string]int64)
        (*mp)["default"] = 180
        return nil
    })
// platform not present -> expect default from map
got := tr.GetTraceDataMaxDurationDay(ctx, stringPtr("missing"))
assert.Equal(t, int64(180), got)
```
- Reference: Go Testing Best Practices; ByteDance 服务端代码规范（测试用例覆盖与回归防护）

CozeLoop (commented):
Aime CR: Focused on trace duration default fallback and deployment config consistency. Added comments for test coverage and confirmed no risk on YAML defaults.

kidkidkid (approved):

PR #276, opened 2025-10-31
[feat][backend]: ob get trace refactor

kidkidkid:
#### Check the PR title.

en:
zh(optional):


backend/modules/observability/application/trace.go
```diff
@@ -1032,3 +1008,57 @@ func (t *TraceApplication) validateExtractSpanInfoReq(ctx context.Context, req *
...
+		return resp
+	}
+	var (
+		g    errgroup.Group
+		lock sync.Mutex
+	)
+	g.Go(func() error {
+		defer goroutine.Recovery(ctx)
+		_, userMap, _ := t.userSvc.GetUserInfo(ctx, req.UserIDs)
+		lock.Lock()
```
CozeLoop: [🚨Should-Fix] Unnecessary locking across goroutines when writing distinct fields
The current implementation uses a shared mutex to guard assignments to different struct fields (UserMap/EvalMap/TagMap). Since these are independent results, a simpler and faster pattern is to collect values in local variables within each goroutine and assign them once after `g.Wait()`. This avoids lock contention and reduces complexity while remaining race-free.

Concrete fix proposal:
```
var (
    g errgroup.Group
    userMap map[string]*commdo.UserInfo
    evalMap map[int64]*rpc.Evaluator
    tagMap  map[int64]*rpc.TagInfo
)
g.Go(func() error { _, userMap, _ = t.userSvc.GetUserInfo(ctx, req.UserIDs); return nil })
g.Go(func() error { _, evalMap, _ = t.evalSvc.BatchGetEvaluatorVersions(ctx, &rpc.BatchGetEvaluatorVersionsParam{WorkspaceID: req.WorkspaceID, EvaluatorVersionIds: req.EvaluatorIDs}); return nil })
g.Go(func() error { tagMap, _ = t.tagSvc.BatchGetTagInfo(ctx, req.WorkspaceID, req.TagKeyIDs); return nil })
_ = g.Wait()
return GetDisplayInfoResponse{UserMap: userMap, EvalMap: evalMap, TagMap: tagMap}
```
References: Effective Go – Concurrency; Go Code Review Comments (Concurrency patterns).

backend/modules/observability/application/trace.go
```diff
@@ -1032,3 +1008,57 @@ func (t *TraceApplication) validateExtractSpanInfoReq(ctx context.Context, req *
...
+	if len(req.UserIDs) == 0 && len(req.EvaluatorIDs) == 0 && len(req.TagKeyIDs) == 0 {
+		return resp
+	}
+	var (
+		g    errgroup.Group
+		lock sync.Mutex
+	)
+	g.Go(func() error {
+		defer goroutine.Recovery(ctx)
+		_, userMap, _ := t.userSvc.GetUserInfo(ctx, req.UserIDs)
```
CozeLoop: [🚨Should-Fix] Errors from RPC calls are ignored in `GetDisplayInfo`
All three goroutines drop returned errors (using `_`). If any call fails, the response silently omits data, making issues hard to diagnose. At minimum, capture and log errors; optionally propagate the first error via `errgroup`.

Concrete fix proposal:
```
g.Go(func() error {
    defer goroutine.Recovery(ctx)
    _, userMap, err := t.userSvc.GetUserInfo(ctx, req.UserIDs)
    if err != nil { logs.CtxWarn(ctx, "GetUserInfo failed: %v", err) }
    resp.UserMap = userMap
    return nil
})
```
Consider `errgroup.WithContext(ctx)` and returning `err` to stop early on systemic failures.
References: Go Code Review Comments – Error handling; Effective Go – Handle errors explicitly.

backend/modules/observability/infra/config/trace.go
```diff
@@ -50,12 +52,29 @@ func (t *TraceConfigCenter) GetPlatformTenants(ctx context.Context) (*config.Pla
...
 	cfg := new(config.SpanTransHandlerConfig)
-	if err := t.UnmarshalKey(ctx, platformSpanHandlerCfgKey, cfg); err != nil {
-		return nil, err
-	}
-	return cfg, nil
+	val := t.Get(ctx, platformSpanHandlerCfgKey)
+	if val == nil {
+		return nil, fmt.Errorf("config not found")
+	}
+	strVal, ok := val.(string)
```
CozeLoop: [💡Suggestion] Avoid type assertion to string + JSON roundtrip for configuration
`GetPlatformSpansTrans` now reads a raw string and re-unmarshals into `SpanTransHandlerConfig`. If the config loader can return a typed copy via `UnmarshalKey`, you can avoid the conversion, reduce allocations, and make type mismatches impossible at runtime.

Concrete fix proposal:
```
cfg := new(config.SpanTransHandlerConfig)
if err := t.UnmarshalKey(ctx, platformSpanHandlerCfgKey, cfg); err != nil { return nil, err }
ret := cfg.PlatformCfg[string(platformType)]
// merge inheritance as you do
```
If the change was intended to avoid shared references, consider returning a fresh struct from the loader or deep-copying the result instead of re-parsing JSON.
References: Effective Go – Data; Performance considerations for JSON unmarshalling.

backend/modules/observability/application/trace.go
```diff
@@ -1032,3 +1008,57 @@ func (t *TraceApplication) validateExtractSpanInfoReq(ctx context.Context, req *
...
+	TagMap  map[int64]*rpc.TagInfo
+}
+
+func (t *TraceApplication) GetDisplayInfo(ctx context.Context, req *GetDisplayInfoRequest) GetDisplayInfoResponse {
+	resp := GetDisplayInfoResponse{}
+	if len(req.UserIDs) == 0 && len(req.EvaluatorIDs) == 0 && len(req.TagKeyIDs) == 0 {
+		return resp
+	}
+	var (
+		g    errgroup.Group
```
CozeLoop: [💡Suggestion] Use `errgroup.WithContext` to propagate cancellation
Using a plain `errgroup.Group` means goroutines continue running even if the parent context is canceled. Switching to `errgroup.WithContext(ctx)` lets you pass the derived ctx to RPC calls and stop work promptly under cancellations.

Concrete fix proposal:
```
g, gctx := errgroup.WithContext(ctx)
g.Go(func() error { _, userMap, err := t.userSvc.GetUserInfo(gctx, req.UserIDs); /* ... */; return nil })
// ... other goroutines
_ = g.Wait()
```
References: Effective Go – Concurrency; x/sync/errgroup docs.

CozeLoop (changes requested):
Aime Review: Identified 1 Must-Fix and several Should-Fix/Suggestions in concurrency and configuration handling. Please see inline comments for details.

taoyifan89 (approved):

JqRrt (approved):

PR #289, opened 2025-11-04
[fix][backend]fix backfill count

zero3233-bd:
#### Check the PR title.

en:
zh(optional):

### DIFF
--- a/backend/modules/observability/domain/task/service/taskexe/tracehub/backfill.go
+++ b/backend/modules/observability/domain/task/service/taskexe/tracehub/backfill.go
@@ -433,6 +433,7 @@ func (h *TraceHubServiceImpl) processSpansForBackfill(ctx context.Context, spans
 			// Continue with the next batch without stopping due to a single failure
 			continue
 		}
+		// ml_flow rate-limited: 50/5s
 		time.Sleep(5 * time.Second)
 	}
 
--- a/backend/modules/observability/domain/task/service/taskexe/processor/auto_evaluate.go
+++ b/backend/modules/observability/domain/task/service/taskexe/processor/auto_evaluate.go
@@ -191,7 +191,7 @@ func (p *AutoEvaluteProcessor) OnCreateTaskChange(ctx context.Context, currentTa
 			case task.TimeUnitWeek:
 				runEndAt = runStartAt + (currentTask.Sampler.CycleInterval)*7*24*time.Hour.Milliseconds()
 			default:
-				runEndAt = runStartAt + (currentTask.Sampler.CycleInterval)*10*time.Minute.Milliseconds()
+				runEndAt = runStartAt + (currentTask.Sampler.CycleInterval)*24*time.Hour.Milliseconds()
 			}
 		}
 		err = p.OnCreateTaskRunChange(ctx, taskexe.OnCreateTaskRunChangeReq{


CozeLoop (changes requested):
Summary: The locking TTL change and pagination token fix look good. However, the newly added per‑batch fixed 5s sleep in `processSpansForBackfill` introduces a performance regression and ignores context cancellation. Please replace with a context‑aware wait (respect `ctx.Done()`) and consider making the delay tunable via configuration. Batch size reduction is likely acceptable if driven by capacity, but should be configurable.

Requesting changes for the context handling fix.

CozeLoop (changes requested):
Summary: Found a P1 issue where non-context-aware `time.Sleep` can stall cancellation and leak goroutines; and a P2 performance/configurability issue due to hard-coded `batchSize=50` with fixed delay. Suggested context-aware sleep via `select` on `ctx.Done()` and making batch size configurable (keep 100 default). Also confirmed page token assignment is correct. Please address the P1 before merge.

backend/modules/observability/domain/task/service/taskexe/tracehub/backfill.go
```diff
@@ -29,6 +29,7 @@ const (
 	pageSize                = 500
 	backfillLockKeyTemplate = "observability:tracehub:backfill:%d"
 	backfillLockMaxHold     = 24 * time.Hour
+	backfillLockTTL         = 3 * time.Minute
```
CozeLoop, line 32: Change has no risk.

backend/modules/observability/domain/task/service/taskexe/tracehub/backfill.go
```diff
@@ -311,7 +312,7 @@ func (h *TraceHubServiceImpl) fetchAndSendSpans(ctx context.Context, listParam *
 			logs.CtxInfo(ctx, "completed listing spans, total_count=%d, task_id=%d", totalCount, sub.t.GetID())
 			break
 		}
-
+		listParam.PageToken = result.PageToken
```
CozeLoop, line 315: Change has no risk.

backend/modules/observability/domain/task/service/taskexe/tracehub/backfill.go
```diff
@@ -417,7 +418,7 @@ func (h *TraceHubServiceImpl) applySampling(spans []*loop_span.Span, sub *spanSu
 // processSpansForBackfill handles spans for backfill
 func (h *TraceHubServiceImpl) processSpansForBackfill(ctx context.Context, spans []*loop_span.Span, sub *spanSubscriber) error {
 	// Batch processing spans for efficiency
-	const batchSize = 100
+	const batchSize = 50
```
CozeLoop, line 421: [P2] Hard-coded batch size (50) may cause unnecessary throughput loss or under-utilization under varying load. Suggest making it configurable (with a sensible default) rather than a fixed constant. For example:

```go
const batchSize = 100 // or read from configuration
```

Reference: Effective Go (code clarity), and internal CozeLoop server spec—prefer configuration over magic numbers.

backend/modules/observability/domain/task/service/taskexe/tracehub/backfill.go
```diff
@@ -432,6 +433,7 @@ func (h *TraceHubServiceImpl) processSpansForBackfill(ctx context.Context, spans
 			// Continue with the next batch without stopping due to a single failure
 			continue
 		}
+		time.Sleep(5 * time.Second)
```
CozeLoop, line 437: [P1] Non-context-aware sleep can stall cancellation and risk leaking goroutines in long backfill runs. Replace with a context-aware wait so the loop exits promptly:

```go
select {
case <-ctx.Done():
    return ctx.Err()
case <-time.After(5 * time.Second):
}
```

This respects `ctx.Done()` and avoids blocking during shutdown. Reference: Effective Go (concurrency and cancellation), internal server spec on context propagation.

backend/modules/observability/domain/task/service/taskexe/tracehub/backfill.go
```diff
@@ -43,7 +44,7 @@ func (h *TraceHubServiceImpl) BackFill(ctx context.Context, event *entity.BackFi
 	)
 	if h.locker != nil && event != nil {
 		lockKey = fmt.Sprintf(backfillLockKeyTemplate, event.TaskID)
-		locked, lockCtx, cancel, lockErr := h.locker.LockWithRenew(ctx, lockKey, transformTaskStatusLockTTL, backfillLockMaxHold)
+		locked, lockCtx, cancel, lockErr := h.locker.LockWithRenew(ctx, lockKey, backfillLockTTL, backfillLockMaxHold)
```
CozeLoop, line 47: Change has no risk.

taoyifan89 (approved):

PR #291, opened 2025-11-05
[fix][backend] transformTaskStatus

zero3233-bd:
#### Check the PR title.

en:
zh(optional):

### DIFF
--- a/backend/modules/observability/domain/task/entity/task.go
+++ b/backend/modules/observability/domain/task/entity/task.go
@@ -127,21 +127,23 @@ type DataReflowRunConfig struct {
 }
 
 func (t ObservabilityTask) GetRunTimeRange() (startAt, endAt int64) {
-	var runStartAt, runEndAt int64
-	runStartAt = t.EffectiveTime.StartAt
+	if t.EffectiveTime == nil {
+		return 0, 0
+	}
+	startAt = t.EffectiveTime.StartAt
 	if !t.Sampler.IsCycle {
-		runEndAt = t.EffectiveTime.EndAt
+		endAt = t.EffectiveTime.EndAt
 	} else {
 		switch t.Sampler.CycleTimeUnit {
 		case task.TimeUnitDay:
-			runEndAt = runStartAt + (t.Sampler.CycleInterval)*24*time.Hour.Milliseconds()
+			endAt = startAt + (t.Sampler.CycleInterval)*24*time.Hour.Milliseconds()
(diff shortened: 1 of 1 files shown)


backend/modules/observability/domain/task/entity/task.go
```diff
@@ -126,6 +126,24 @@ type DataReflowRunConfig struct {
...
+		switch t.Sampler.CycleTimeUnit {
+		case task.TimeUnitDay:
+			runEndAt = runStartAt + (t.Sampler.CycleInterval)*24*time.Hour.Milliseconds()
+		case task.TimeUnitWeek:
+			runEndAt = runStartAt + (t.Sampler.CycleInterval)*7*24*time.Hour.Milliseconds()
+		default:
+			runEndAt = runStartAt + (t.Sampler.CycleInterval)*24*time.Hour.Milliseconds()
+		}
+	}
+	return runStartAt, runEndAt
```
CozeLoop: [P1] Nil pointer risk in `GetRunTimeRange`

Problem: This method dereferences `t.EffectiveTime` and `t.Sampler` without nil checks. If either is nil (e.g., tasks created without a sampler or with missing effective time), it will panic. `ShouldTriggerNewData` only guards `EffectiveTime != nil` and does not validate `Sampler`, and `transformTaskStatus` calls this method directly.

Why it matters: These fields are declared as pointers in `ObservabilityTask`; defensive handling is required to avoid runtime panics on production tasks with incomplete configuration.

Fix suggestion (treat nil sampler as non‑cycle; short‑circuit when `EffectiveTime` is nil):
```go
func (t ObservabilityTask) GetRunTimeRange() (startAt, endAt int64) {
    if t.EffectiveTime == nil { return 0, 0 }
    startAt = t.EffectiveTime.StartAt
    if t.Sampler == nil || !t.Sampler.IsCycle { return startAt, t.EffectiveTime.EndAt }
    switch t.Sampler.CycleTimeUnit {
    case task.TimeUnitDay:
        endAt = startAt + t.Sampler.CycleInterval*24*time.Hour.Milliseconds()
    case task.TimeUnitWeek:
        endAt = startAt + t.Sampler.CycleInterval*7*24*time.Hour.Milliseconds()
    default:
        endAt = startAt + t.Sampler.CycleInterval*24*time.Hour.Milliseconds()
    }
    return startAt, endAt
}
```

taoyifan89 (approved):

PR #298, opened 2025-11-10
[feat][backend] ob trans refactor

kidkidkid:
#### Check the PR title.

en:
zh(optional):

### DIFF
--- a/backend/modules/observability/domain/trace/service/trace_service.go
+++ b/backend/modules/observability/domain/trace/service/trace_service.go
@@ -573,15 +573,6 @@ func (r *TraceServiceImpl) GetTracesAdvanceInfo(ctx context.Context, req *GetTra
 	resp := &GetTracesAdvanceInfoResp{
 		Infos: []*loop_span.TraceAdvanceInfo{},
 	}
-	// use one processor...
-	processors, err := r.buildHelper.BuildAdvanceInfoProcessors(ctx, span_processor.Settings{
-		WorkspaceId:  req.WorkspaceID,
-		PlatformType: req.PlatformType,
-	})
-	if err != nil {
-		logs.CtxError(ctx, "Fail to build advance info processor, %v", err)
-		return nil, err
-	}
 	for _, v := range req.Traces {
 		g.Go(func() error {
 			defer goroutine.Recovery(ctx)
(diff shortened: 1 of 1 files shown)


backend/modules/observability/domain/trace/entity/loop_span/span.go
```diff
@@ -550,8 +550,11 @@ func (s *Span) ClipSpan() {
 }
 
 func (s SpanList) Stat(ctx context.Context) (inputTokens, outputTokens int64, err error) {
-	modelSpans := s.FilterModelSpans()
-	for _, v := range modelSpans {
+	filter := GetModelSpansFilter()
+	for _, v := range s {
+		if !filter.Satisfied(v) {
+			continue
+		}
```
CozeLoop, line 557: [Suggestion] Redundant filtering: `Stat` applies `GetModelSpansFilter()` again even when the caller has already filtered spans at the repository level (e.g., `GetTracesAdvanceInfo` sets `Filters: GetModelSpansFilter()` in the query). This double filtering adds overhead and can drift if the predicate evolves. Consider either computing tokens over the provided spans directly, or add a `StatWithFilter(ctx, f *FilterFields)` helper so callers can control filtering explicitly.

Example:
```go
func (s SpanList) Stat(ctx context.Context) (int64, int64, error) {
    var inTotal, outTotal int64
    for _, v := range s {
        in, out, err := v.getTokens(ctx)
        if err != nil { return -1, -1, err }
        inTotal += in; outTotal += out
    }
    return inTotal, outTotal, nil
}
```

backend/modules/observability/domain/trace/entity/loop_span/trans.go
```diff
@@ -128,15 +130,17 @@ func (c *OutputFilter) transform(ctx context.Context, span *Span) {
...
+		if c.TagFilter != nil {
+			c.TagFilter.keyBlackListMap = slices.ToMap(c.TagFilter.KeyBlackList, func(e string) (string, bool) { return e, true })
+		}
+		if c.InputFilter != nil {
+			c.InputFilter.keyWhiteListMap = slices.ToMap(c.InputFilter.KeyWhiteList, func(e string) (string, bool) { return e, true })
+		}
+		if c.OutputFilter != nil {
+			c.OutputFilter.keyWhiteListMap = slices.ToMap(c.OutputFilter.KeyWhiteList, func(e string) (string, bool) { return e, true })
+		}
+	})
```
CozeLoop, line 143: [Should-Fix] `sync.Once` here makes `SpanTransConfig` non-reloadable: if tag/input/output lists are updated at runtime (e.g., via dynamic config), the maps won’t be rebuilt because `init()` will be skipped after the first call. This can leave stale filters in effect. Consider reconstructing new `SpanTransConfig` instances on updates, or provide a `Reset()` that zeroes `initOnce` and map caches to allow re-init. Also ensure the struct is not copied after init—copying `sync.Once` post-use is undefined.

Example:
```go
func (c *SpanTransConfig) Reset() {
    c.initOnce = sync.Once{}
    if c.TagFilter != nil { c.TagFilter.keyBlackListMap = nil }
    if c.InputFilter != nil { c.InputFilter.keyWhiteListMap = nil }
    if c.OutputFilter != nil { c.OutputFilter.keyWhiteListMap = nil }
}
```

backend/modules/observability/domain/trace/service/trace_service.go
```diff
@@ -604,6 +596,19 @@ func (r *TraceServiceImpl) GetTracesAdvanceInfo(ctx context.Context, req *GetTra
...
+				PlatformType:    req.PlatformType,
+				QueryStartTime:  v.StartTime,
+				QueryEndTime:    v.EndTime + defaultTimeRange,
+				SpanDoubleCheck: true,
+				QueryTenants:    tenants,
+				QueryTraceID:    v.TraceID,
+			})
+			if err != nil {
+				logs.CtxError(ctx, "Fail to build advance info processor, %v", err)
+				return err
```
CozeLoop, line 610: [Suggestion] Processor construction per trace could be costly if factories are heavy. Building processors inside each goroutine allows per-trace settings (start/end time, tenants), but if processor creation is stateless and does not depend on per-trace values, consider caching/reusing them outside the loop to reduce allocations and CPU. Alternatively, pass per-trace context as parameters to `Transform` while reusing the same processor instances.

PR #303, opened 2025-11-11
[feat][backend] trace oapi support original tag

cuichen01:
#### Check the PR title.

en:
zh(optional):

### DIFF
--- a/backend/modules/observability/application/convertor/trace/span_test.go
+++ b/backend/modules/observability/application/convertor/trace/span_test.go
@@ -324,7 +324,7 @@ func TestFilterFieldsDTO2DO(t *testing.T) {
 func TestFilterFieldListDTO2DO(t *testing.T) {
 	list := FilterFieldListDTO2DO([]*filter.FilterField{
 		nil,
-		&filter.FilterField{
+		{
 			FieldName: ptr.Of("name"),
 		},
 	})


backend/modules/observability/application/convertor/trace/span.go
```diff
@@ -36,6 +37,16 @@ func SpanDO2DTO(
 		Output:          s.Output,
 		LogicDeleteDate: ptr.Of(time_util.MicroSec2MillSec(s.LogicDeleteTime)), // to ms
 	}
+	if needOriginalTags {
```
CozeLoop, line 41: [🚨Should-Fix] Avoid sharing mutable maps when exposing original tags

Assigning `s.SystemTagsString/Long/Double` and `Tags*` maps directly to `outSpan` keeps references to the underlying DO maps. If those maps are mutated later or reused across requests, responses may observe inconsistent values and this can invite data races under concurrent handlers. Safer to copy the maps before setting them.

Concrete suggestion:
```go
outSpan.SystemTagsString = make(map[string]string, len(s.SystemTagsString))
for k, v := range s.SystemTagsString { outSpan.SystemTagsString[k] = v }

outSpan.SystemTagsLong = make(map[string]int64, len(s.SystemTagsLong))
for k, v := range s.SystemTagsLong { outSpan.SystemTagsLong[k] = v }
// Repeat for SystemTagsDouble, TagsString, TagsLong, TagsDouble, TagsBool, TagsBytes
```

backend/modules/observability/application/openapi.go
```diff
@@ -514,7 +514,7 @@ func (o *OpenAPIApplication) SearchTraceOApi(ctx context.Context, req *openapi.S
 	logs.CtxInfo(ctx, "SearchTrace successfully, spans count %d", len(sResp.Spans))
 	return &openapi.SearchTraceOApiResponse{
 		Data: &openapi.SearchTraceOApiData{
-			Spans: tconv.SpanListDO2DTO(sResp.Spans, nil, nil, nil),
+			Spans: tconv.SpanListDO2DTO(sResp.Spans, nil, nil, nil, req.GetNeedOriginalTags()),
```
CozeLoop, line 517: [💡Suggestion] Bound payload by limiting original tag maps when enabled

Forwarding `req.GetNeedOriginalTags()` will include all raw tag maps in responses. For large traces this can significantly increase payload size and memory allocations and may inadvertently expose internal keys. Consider adding caps/whitelists in the converter when `needOriginalTags` is true.

Example idea:
```go
const maxKeys = 200
func clampStringMap(m map[string]string, n int) map[string]string {
    if len(m) <= n { return m }
    out := make(map[string]string, n)
    i := 0
    for k, v := range m {
        out[k] = v
        i++
        if i >= n { break }
    }
    return out
}
// apply clamp before setting on outSpan
```

PR #311, opened 2025-11-17
[feat][backend]support feedback auto task

zero3233-bd:
#### Check the PR title

en:
zh(optional):


backend/modules/observability/infra/repo/task.go
```diff
@@ -324,7 +315,7 @@ func (v *TaskRepoImpl) IncrTaskRunFailCount(ctx context.Context, taskID, taskRun
 	return v.TaskRunRedisDao.IncrTaskRunFailCount(ctx, taskID, taskRunID, time.Duration(ttl)*time.Second)
```
CozeLoop: 🚨 SHOULD FIX: Fail‑count TTL uses seconds but `ttl` is in milliseconds. Use `time.Millisecond` to avoid inflated retention windows.

Suggested fix:
```go
return v.TaskRunRedisDao.IncrTaskRunFailCount(ctx, taskID, taskRunID, time.Duration(ttl)*time.Millisecond)
```

PR #319, opened 2025-11-24
[fix][evaluation] fix the evaluator ecosystem

tpfz:
#### Check the PR title

en:
zh(optional):

### DIFF
--- a/backend/modules/evaluation/application/evaluator_app_test.go
+++ b/backend/modules/evaluation/application/evaluator_app_test.go
@@ -867,6 +867,195 @@ func TestEvaluatorHandlerImpl_GetEvaluatorVersion(t *testing.T) {
 	}
 }
 
+// TestEvaluatorHandlerImpl_BatchGetEvaluatorVersions 测试 BatchGetEvaluatorVersions 方法
+func TestEvaluatorHandlerImpl_BatchGetEvaluatorVersions(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+
+	mockAuth := rpcmocks.NewMockIAuthProvider(ctrl)
+	mockEvaluatorService := mocks.NewMockEvaluatorService(ctrl)
+	mockUserInfoService := userinfomocks.NewMockUserInfoService(ctrl)
+	mockConfiger := confmocks.NewMockIConfiger(ctrl)
+
+	app := &EvaluatorHandlerImpl{
+		auth:             mockAuth,
+		evaluatorService: mockEvaluatorService,
+		userInfoService:  mockUserInfoService,
+		configer:         mockConfiger,
+	}
+
+	workspaceID := int64(100)
+	builtinSpaceID := int64(200)
+	versionIDs := []int64{11, 22}
+
+	workspaceEvaluator := &entity.Evaluator{
+		ID:            1,
+		SpaceID:       workspaceID,
+		EvaluatorType: entity.EvaluatorTypePrompt,
+		PromptEvaluatorVersion: &entity.PromptEvaluatorVersion{
+			ID:          versionIDs[0],
+			EvaluatorID: 1,
+			Version:     "v1",
+		},
+	}
+	builtinEvaluator := &entity.Evaluator{
+		ID:            2,
+		SpaceID:       builtinSpaceID,
+		Builtin:       true,
+		EvaluatorType: entity.EvaluatorTypePrompt,
+		PromptEvaluatorVersion: &entity.PromptEvaluatorVersion{
+			ID:          versionIDs[1],
+			EvaluatorID: 2,
+			Version:     "v2",
+		},
+	}
+
+	tests := []struct {
+		name               string
+		req                *evaluatorservice.BatchGetEvaluatorVersionsRequest
+		mockSetup          func()
+		wantErr            bool
+		wantErrCode        int32
+		wantEvaluatorCount int
+	}{
+		{
+			name: "success - workspace evaluators only",
+			req: &evaluatorservice.BatchGetEvaluatorVersionsRequest{
+				WorkspaceID:         workspaceID,
+				EvaluatorVersionIds: []int64{versionIDs[0]},
+				IncludeDeleted:      gptr.Of(false),
+			},
+			mockSetup: func() {
+				mockEvaluatorService.EXPECT().
+					BatchGetEvaluatorVersion(gomock.Any(), (*int64)(nil), []int64{versionIDs[0]}, false).
+					Return([]*entity.Evaluator{workspaceEvaluator}, nil)
+
+				mockAuth.EXPECT().
+					Authorization(gomock.Any(), &rpc.AuthorizationParam{
+						ObjectID:      strconv.FormatInt(workspaceID, 10),
+						SpaceID:       workspaceID,
+						ActionObjects: []*rpc.ActionObject{{Action: gptr.Of("listLoopEvaluator"), EntityType: gptr.Of(rpc.AuthEntityType_Space)}},
+					}).
+					Return(nil)
+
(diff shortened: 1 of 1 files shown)


HearyShen (dismissed):

backend/modules/evaluation/application/evaluator_app.go
```diff
@@ -665,21 +665,42 @@ func (e *EvaluatorHandlerImpl) GetEvaluatorVersion(ctx context.Context, request
...
+				ActionObjects: []*rpc.ActionObject{{Action: gptr.Of("listLoopEvaluator"), EntityType: gptr.Of(rpc.AuthEntityType_Space)}},
+			})
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			// 如果不是，使用预置空间校验的 authBuiltinManagement 函数进行校验
+			err = e.authBuiltinManagement(ctx, spaceID, spaceTypeBuiltin, false)
+			if err != nil {
+				return nil, err
```
CozeLoop, line 701: [MAJOR] Authorization path applies builtin management to all non-request spaces
The else branch calls `authBuiltinManagement` for ANY `spaceID != request.WorkspaceID`. This assumes those evaluators belong to builtin spaces, which is not guaranteed. For non‑builtin evaluators in other spaces, this will apply the wrong auth policy and may either over‑grant or deny access.

Why it’s a problem: Authorization must depend on the evaluator’s ownership type (builtin vs normal). Using builtin management checks for arbitrary spaces creates security and correctness issues when cross‑space data is involved.

Proposal: Drive auth by the evaluator’s `Builtin` flag. For builtin evaluators, use builtin management; for normal ones, perform space‑level `Authorization` against that evaluator’s space. Example:

```go
// group unique spaceIDs per type
for _, evaluatorDO := range evaluatorDOList {
    spaceID := evaluatorDO.SpaceID
    if checkedSpaceIDs[spaceID] { continue }
    checkedSpaceIDs[spaceID] = true

    if evaluatorDO.Builtin { // builtin space
        if err = e.authBuiltinManagement(ctx, spaceID, spaceTypeBuiltin, false); err != nil { return nil, err }
    } else { // normal space
        if err = e.auth.Authorization(ctx, &rpc.AuthorizationParam{
            ObjectID:      strconv.FormatInt(spaceID, 10),
            SpaceID:       spaceID,
            ActionObjects: []*rpc.ActionObject{{Action: gptr.Of("listLoopEvaluator"), EntityType: gptr.Of(rpc.AuthEntityType_Space)}},
        }); err != nil { return nil, err }
    }
}
```

Reference: ByteDance Go service code spec – authorization must match resource ownership; do not mix builtin management with normal space ACLs.

backend/modules/evaluation/application/evaluator_app.go
```diff
@@ -665,21 +665,42 @@ func (e *EvaluatorHandlerImpl) GetEvaluatorVersion(ctx context.Context, request
...
-		ActionObjects: []*rpc.ActionObject{{Action: gptr.Of("listLoopEvaluator"), EntityType: gptr.Of(rpc.AuthEntityType_Space)}},
-	})
-	if err != nil {
-		return nil, err
+	// 按 SpaceID 分组进行权限校验
+	requestWorkspaceID := request.WorkspaceID
+	checkedSpaceIDs := make(map[int64]bool)
+	for _, evaluatorDO := range evaluatorDOList {
+		spaceID := evaluatorDO.SpaceID
+		// 如果已经校验过该空间，跳过
```
CozeLoop, line 681: [MINOR] Add a nil guard for evaluator items to avoid unexpected panic
`spaceID := evaluatorDO.SpaceID` will panic if the service ever returns a `nil` element in the list (e.g., merged from heterogeneous queries). Defensive programming here reduces blast radius.

Suggestion:

```go
for _, evaluatorDO := range evaluatorDOList {
    if evaluatorDO == nil { continue }
    spaceID := evaluatorDO.SpaceID
    // ...
}
```

Rationale: Robustness against edge cases aligns with Effective Go and internal code specs (avoid panics in handlers).

backend/modules/evaluation/domain/entity/evaluator_version_custom_rpc.go
```diff
@@ -15,6 +15,13 @@ import (
 	"github.com/coze-dev/coze-loop/backend/pkg/errorx"
 )
 
+// EvaluatorAccessProtocol 评估器接入协议类型
+type EvaluatorAccessProtocol = string
+
+const (
+	EvaluatorAccessProtocolRPC = "rpc"
+)
```
CozeLoop, line 23: [P2] Enforce allowed values for `EvaluatorAccessProtocol` at domain layer
The new alias and constant (`EvaluatorAccessProtocolRPC = "rpc"`) are good, but the domain currently only checks non-empty values. To align with the IDL enum and prevent misconfiguration, consider centralizing a validator (e.g., `IsValidEvaluatorAccessProtocol`) and invoke it during `ValidateBaseInfo`.

Example:

```go
func IsValidEvaluatorAccessProtocol(p EvaluatorAccessProtocol) bool {
    switch p {
    case EvaluatorAccessProtocolRPC:
        return true
    default:
        return false
    }
}
```

This ensures only supported protocols are accepted now and makes future extensions explicit.

backend/modules/evaluation/application/evaluator_app.go
```diff
@@ -665,21 +665,42 @@ func (e *EvaluatorHandlerImpl) GetEvaluatorVersion(ctx context.Context, request
...
+		checkedSpaceIDs[spaceID] = true
+
+		// 如果是请求的空间下的，使用原来的权限校验逻辑
+		if spaceID == requestWorkspaceID {
+			err = e.auth.Authorization(ctx, &rpc.AuthorizationParam{
+				ObjectID:      strconv.FormatInt(spaceID, 10),
+				SpaceID:       spaceID,
+				ActionObjects: []*rpc.ActionObject{{Action: gptr.Of("listLoopEvaluator"), EntityType: gptr.Of(rpc.AuthEntityType_Space)}},
+			})
+			if err != nil {
```
CozeLoop, line 694: [P1] Missing workspace-level list permission check for builtin case (parity with GetEvaluatorVersion)
In `GetEvaluatorVersion`, builtin flow performs two checks: (1) builtin management on the evaluator’s space, and (2) workspace-level `listLoopEvaluator` Authorization on the requesting space. Here, for builtin spaces `!= request.WorkspaceID`, only the builtin management check is performed. The workspace-level check is missing, which reduces parity and may bypass intended ACL.

Consider adding the second check for the requesting workspace when any builtin evaluator is included:

```go
// after passing builtin management for a builtin evaluator
if err = e.auth.Authorization(ctx, &rpc.AuthorizationParam{
    ObjectID:      strconv.FormatInt(requestWorkspaceID, 10),
    SpaceID:       requestWorkspaceID,
    ActionObjects: []*rpc.ActionObject{{Action: gptr.Of("listLoopEvaluator"), EntityType: gptr.Of(rpc.AuthEntityType_Space)}},
}); err != nil { return nil, err }
```

This keeps behavior consistent and prevents privilege gaps.

CozeLoop (changes requested):
Overall this PR improves type clarity (EvaluatorAccessProtocol) and broadens BatchGetEvaluatorVersions to fetch across spaces. However, the new auth flow introduces correctness risks:

1) All non-request spaces are treated as builtin via authBuiltinManagement. Authorization must be driven by evaluator ownership (builtin vs normal), otherwise policy mismatches or privilege gaps can occur.
2) For builtin evaluators in other spaces, parity with GetEvaluatorVersion is missing: the workspace-level list permission check should also be performed.
3) Minor robustness: add a nil guard in the iteration to avoid panic on unexpected nil items.

Please address these before merging.

PR #327, opened 2025-12-04
[fix][evaluation]eval online

tpfz:
#### Check the PR title

en:
zh(optional):

### DIFF
--- a/backend/modules/evaluation/domain/service/expt_manage_execution_validation_test.go
+++ b/backend/modules/evaluation/domain/service/expt_manage_execution_validation_test.go
@@ -115,24 +115,43 @@ func TestExptMangerImpl_checkTargetConnector_VersionIDValidation(t *testing.T) {
 			expectedError: false,
 		},
 		{
-			name: "loop_trace_target_should_skip_validation",
+			name: "loop_trace_target_should_fail_validation", // 修正测试名称以符合实际行为
 			expt: &entity.Experiment{
 				TargetVersionID: 123,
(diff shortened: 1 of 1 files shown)


backend/modules/evaluation/domain/service/expt_manage_execution_impl.go
```diff
@@ -120,8 +120,7 @@ func (e *ExptMangerImpl) CheckConnector(ctx context.Context, expt *entity.Experi
 }
 
 func (e *ExptMangerImpl) checkTargetConnector(ctx context.Context, expt *entity.Experiment, session *entity.Session) error {
-	if expt.Target == nil ||
-		expt.Target.EvalTargetType == entity.EvalTargetTypeLoopTrace {
+	if expt.Target == nil || expt.ExptType == entity.ExptType_Online {
```
CozeLoop, line 123: This change from `expt.Target.EvalTargetType == entity.EvalTargetTypeLoopTrace` to `expt.ExptType == entity.ExptType_Online` appears to be incorrect. It assumes that all online experiments should skip the target connector check, which may not be true. The original logic correctly skipped the check for `LoopTrace` targets, which likely do not require a connector. This change could introduce a bug where connector validation is skipped for online experiments that do require it. The logic should likely be a combination of both checks, not a replacement.

CozeLoop (changes requested):
Found a critical issue that needs to be addressed before merging.

HearyShen (approved):

PR #418, opened 2026-02-04
[feat][evaluation]open api evaluator

tpfz:
#### Check the PR title

en:
zh(optional):

### DIFF
--- a/backend/modules/evaluation/domain/service/expt_manage_execution_impl_test.go
+++ b/backend/modules/evaluation/domain/service/expt_manage_execution_impl_test.go
@@ -2597,6 +2597,70 @@ func TestExptMangerImpl_checkTargetConnector_WithRuntimeParam(t *testing.T) {
 			wantErr: false,
 		},
 		{
+			name: "TargetType为0时从Target回填并校验runtime_param",
+			expt: &entity.Experiment{
+				ID:              1,
+				TargetVersionID: 1,
+				TargetType:      0, // 未设置，需从 Target 回填
+				Target: &entity.EvalTarget{
+					EvalTargetType: 0,
+					EvalTargetVersion: &entity.EvalTargetVersion{
+						EvalTargetType: entity.EvalTargetTypeLoopPrompt,
+						OutputSchema:   []*entity.ArgsSchema{{Key: gptr.Of("output_field")}},
+					},
+				},
+				EvalSet: &entity.EvaluationSet{
+					EvaluationSetVersion: &entity.EvaluationSetVersion{
+						EvaluationSetSchema: &entity.EvaluationSetSchema{
+							FieldSchemas: []*entity.FieldSchema{{Name: "input_field"}},
+						},
+					},
+				},
+				EvalConf: &entity.EvaluationConfiguration{
+					ConnectorConf: entity.Connector{
+						TargetConf: &entity.TargetConf{
+							TargetVersionID: 1,
+							IngressConf: &entity.TargetIngressConf{
+								EvalSetAdapter: &entity.FieldAdapter{
+									FieldConfs: []*entity.FieldConf{{FromField: "input_field"}},
+								},
+								CustomConf: &entity.FieldAdapter{
+									FieldConfs: []*entity.FieldConf{
+										{
+											FieldName: consts.FieldAdapterBuiltinFieldNameRuntimeParam,
+											Value:     `{"model_config":{"model_id":"fallback_model"}}`,
+										},
+									},
+								},
+							},
+						},
+						EvaluatorsConf: &entity.EvaluatorsConf{
+							EvaluatorConf: []*entity.EvaluatorConf{
+								{
+									EvaluatorVersionID: 1,
+									IngressConf: &entity.EvaluatorIngressConf{
+										EvalSetAdapter: &entity.FieldAdapter{
+											FieldConfs: []*entity.FieldConf{{FromField: "input_field"}},
+										},
+										TargetAdapter: &entity.FieldAdapter{
+											FieldConfs: []*entity.FieldConf{{FromField: "output_field"}},
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			setup: func() {
+				mgr.evalTargetService.(*svcMocks.MockIEvalTargetService).
+					EXPECT().
+					ValidateRuntimeParam(ctx, entity.EvalTargetTypeLoopPrompt, `{"model_config":{"model_id":"fallback_model"}}`).
+					Return(nil)
+			},
+			wantErr: false,
+		},
+		{
 			name: "invalid_runtime_param_format_error",
 			expt: &entity.Experiment{
 				ID:              1,
--- a/backend/modules/evaluation/domain/service/expt_manage_impl_test.go
+++ b/backend/modules/evaluation/domain/service/expt_manage_impl_test.go
@@ -294,6 +294,112 @@ func TestExptMangerImpl_CreateExpt(t *testing.T) {
 	})
 }
 
+func TestExptMangerImpl_CreateExpt_WithExistingTarget(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	defer ctrl.Finish()
+	mgr := newTestExptManager(ctrl)
+	ctx := context.Background()
+	session := &entity.Session{UserID: "1"}
+	targetID := int64(100)
+	targetVersionID := int64(101)
+	param := &entity.CreateExptParam{
+		WorkspaceID:         1,
+		Name:                "expt_with_existing_target",
+		EvalSetID:           2,
+		EvalSetVersionID:    3,
+		TargetID:            &targetID,
+		TargetVersionID:     targetVersionID,
+		EvaluatorVersionIds: []int64{10},
+		ExptConf: &entity.EvaluationConfiguration{
+			ConnectorConf: entity.Connector{
+				TargetConf: &entity.TargetConf{
+					TargetVersionID: targetVersionID,
+					IngressConf: &entity.TargetIngressConf{
+						EvalSetAdapter: &entity.FieldAdapter{
+							FieldConfs: []*entity.FieldConf{{FromField: "field1"}},
+						},
+					},
+				},
+				EvaluatorsConf: &entity.EvaluatorsConf{
+					EvaluatorConf: []*entity.EvaluatorConf{
+						{
+							EvaluatorVersionID: 10,
+							IngressConf: &entity.EvaluatorIngressConf{
+								EvalSetAdapter: &entity.FieldAdapter{
+									FieldConfs: []*entity.FieldConf{{FromField: "field1"}},
+								},
+								TargetAdapter: &entity.FieldAdapter{FieldConfs: []*entity.FieldConf{}},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+
+	mgr.evalTargetService.(*svcMocks.MockIEvalTargetService).
+		EXPECT().
+		GetEvalTargetVersion(ctx, int64(1), targetVersionID, true).
+		Return(&entity.EvalTarget{
+			ID:             targetID,
+			EvalTargetType: 0,
+			EvalTargetVersion: &entity.EvalTargetVersion{
+				ID:             targetVersionID,
+				EvalTargetType: entity.EvalTargetTypeLoopTrace,
+				OutputSchema:   []*entity.ArgsSchema{},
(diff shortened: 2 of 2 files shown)


CozeLoop (commented):
Automated review by Aime: adding line-level comments for identified issues.

backend/modules/evaluation/application/eval_openapi_app.go
```diff
@@ -1000,3 +1010,907 @@ func (e *EvalOpenAPIApplication) GetExperimentAggrResultOApi(ctx context.Context
 		},
 	}, nil
 }
+
+func (e *EvalOpenAPIApplication) ListEvaluatorsOApi(ctx context.Context, req *openapi.ListEvaluatorsOApiRequest) (r *openapi.ListEvaluatorsOApiResponse, err error) {
+	startTime := time.Now().UnixNano() / int64(time.Millisecond)
+	defer func() {
+		e.metric.EmitOpenAPIMetric(ctx, req.GetWorkspaceID(), 0, kitexutil.GetTOMethod(ctx), startTime, err)
```
CozeLoop, line 1016: [Must-Fix] Deferred metric uses `req` without nil guard, causing panic when `req` is nil

The deferred metric logs call `req.GetWorkspaceID()` unconditionally. If this handler returns early because `req == nil`, the defer will still run and panic with a nil pointer dereference, and the client will see a 500 instead of the intended `CommonInvalidParamCode`.

Consider guarding `req` inside the defer (or capturing the workspace ID after the nil check). For example:

```suggestion
startTime := time.Now().UnixNano() / int64(time.Millisecond)
defer func() {
    workspaceID := int64(0)
    if req != nil {
        workspaceID = req.GetWorkspaceID()
    }
    e.metric.EmitOpenAPIMetric(ctx, workspaceID, 0, kitexutil.GetTOMethod(ctx), startTime, err)
}()
```

The same pattern exists in other handlers in this file that use `req` inside the metrics defer (e.g. `BatchGetEvaluatorsOApi`, `CreateExptTemplateOApi`, `SubmitExptFromTemplateOApi`, etc.); they should be updated similarly.

idl/thrift/coze/loop/evaluation/domain_openapi/evaluator.thrift
```diff
@@ -55,11 +87,12 @@ struct EvaluatorVersion {
...
-    3: optional string description
-    4: optional EvaluatorType evaluator_type
-    5: optional bool is_draft_submitted
-    6: optional string latest_version
+    2: optional i64 workspace_id (api.js_conv = 'true', go.tag = 'json:"workspace_id"')
+    3: optional string name
+    4: optional string description
+    5: optional EvaluatorType evaluator_type
+    6: optional bool is_draft_submitted
+    7: optional string latest_version
```
CozeLoop, line 95: [Must-Fix] Changing existing field IDs in `Evaluator` breaks Thrift/IDL compatibility

The previous version of `Evaluator` used IDs 1..6 for `id`, `name`, `description`, `evaluator_type`, `is_draft_submitted` and `latest_version`. This change inserts `workspace_id` as field `2` and shifts the existing fields down to `3..7`. That violates the IDL spec (existing struct fields must not change ID or type) and will break any consumers that rely on the old wire format (Kitex-generated structs, JSON bindings, persistence, etc.).

A safer approach is to keep all existing field IDs unchanged and add `workspace_id` as a new field with a **new** ID.

For example:

```thrift
struct Evaluator {
    1: optional i64 id (api.js_conv = 'true', go.tag = 'json:"id"')
    2: optional string name
    3: optional string description
    4: optional EvaluatorType evaluator_type
    5: optional bool is_draft_submitted
    6: optional string latest_version

    7: optional i64 workspace_id (api.js_conv = 'true', go.tag = 'json:"workspace_id"')

    20: optional EvaluatorVersion current_version
    100: optional common.BaseInfo base_info
}
```

After adjusting the IDL, regenerate the Kitex/domain_openapi code and make sure the Go convertors (`OpenAPIEvaluatorDO2DTO` / `OpenAPIEvaluatorDTO2DO`) continue to map `workspace_id` from/to `SpaceID`. This keeps the wire format backwards compatible while still exposing the workspace dimension via the new field.

Priority: Must-Fix

backend/modules/evaluation/application/convertor/common/openapi.go
```diff
@@ -4,59 +4,442 @@
...
 		Name:      info.Name,
 		AvatarURL: info.AvatarURL,
 		UserID:    info.UserID,
 		Email:     info.Email,
 	}
 }
 
-func OpenAPIArgsSchemaDO2DTOs(dos []*commonentity.ArgsSchema) []*commondto.ArgsSchema {
-	if len(dos) == 0 {
+// OpenAPIContentTypeDO2DTO entity ContentType（首字母大写，如 Text）-> OpenAPI 小写（如 text），DB 存大写
```
CozeLoop, line 39: [Should-Fix] Invalid `rate_limit.period` surfaces as generic internal error instead of validation error

`OpenAPIRateLimitDTO2DO` returns the raw `time.ParseDuration` error. Callers such as `OpenAPIEvaluatorContentDTO2DO` just bubble it up, and the HTTP handler wraps it via `c.Error(err)` without mapping it to a Biz status code. For bad input like `"period": "foo"`, clients will likely see a generic 500 instead of a client-side validation error.

Consider mapping this to a typed invalid-parameter error so that OpenAPI callers get a clear 4xx response. For example, in `OpenAPIEvaluatorContentDTO2DO`:

```go
rateLimit, err := common_convertor.OpenAPIRateLimitDTO2DO(c.RateLimit)
if err != nil {
    return nil, errorx.NewByCode(errno.CommonInvalidParamCode,
        errorx.WithExtraMsg(fmt.Sprintf("invalid rate_limit.period: %s", err.Error())))
}
res.CustomRPCEvaluatorVersion.RateLimit = rateLimit
```

You can move the `errorx` mapping into a small helper if you prefer to keep the pure convertor free of business error types, but the key point is to convert parse failures into a structured invalid-param error instead of a generic internal error.

Priority: Should-Fix

backend/modules/evaluation/application/convertor/experiment/openapi.go
```diff
@@ -1322,3 +1351,606 @@ func OpenAPICustomEvalTargetDO2DTO(do *entity.CustomEvalTarget) *openapiEvalTarg
...
+
+	// Filters
+	filters := dto.Filters
+	if filters == nil || len(filters.GetFilterConditions()) == 0 {
+		if result.FuzzyName == "" && !result.Includes.IsValid() && !result.Excludes.IsValid() {
+			return nil
+		}
+		return result
+	}
+	if filters.LogicOp != nil && strings.ToLower(*filters.LogicOp) != "and" {
```
CozeLoop, line 1911: [Should-Fix] `ExperimentTemplateFilter` silently drops filters and fuzzy-name when `logic_op` != "and"

This converter currently returns `nil` whenever `filters.LogicOp` is set to something other than `"and"`:

```go
if filters.LogicOp != nil && strings.ToLower(*filters.LogicOp) != "and" {
    return nil
}
```

That means a request like

```json
{
  "keyword_search": { "keyword": "foo" },
  "filters": { "logic_op": "or", ... }
}
```

will silently fall back to "no filter at all": both the filter conditions and the previously computed `FuzzyName` are discarded. From the caller’s perspective the API just ignores their filter instead of either applying a reasonable fallback or returning a validation error.

A safer pattern is:

* always preserve `FuzzyName` even if the filter tree is partially invalid; and
* either
  * treat unsupported `logic_op` values as `"and"` (best-effort), **or**
  * reject the request with a clear invalid-parameter error.

For example, if you want a best-effort implementation:

```go
// If logic_op is not specified or unknown, default to AND instead of dropping the filter.
if filters.LogicOp != nil && strings.ToLower(*filters.LogicOp) != "and" {
    // TODO: optionally log/metric a warning here
}
```

If you prefer strict validation, have `OpenAPIExptTemplateFilterDTO2DO` return a typed `error` (e.g. `errorx.NewByCode(errno.CommonInvalidParamCode, ...)`) instead of `nil`, and surface that back to the OpenAPI handler so the client gets a 4xx with a clear message.

Priority: Should-Fix

backend/modules/evaluation/application/convertor/experiment/expt_template.go
```diff
@@ -733,6 +733,66 @@ func buildTemplateScoreWeightConfigDTO(template *entity.ExptTemplate) *domain_ex
...
 
+// TemplateToSubmitExperimentRequest 将实验模板转换为 SubmitExperimentRequest，用于根据模板提交实验
+func TemplateToSubmitExperimentRequest(template *entity.ExptTemplate, name string, workspaceID int64) *expt.SubmitExperimentRequest {
+	if template == nil {
+		return nil
+	}
+	req := &expt.SubmitExperimentRequest{
+		WorkspaceID:    workspaceID,
+		Name:           gptr.Of(name),
+		ExptTemplateID: gptr.Of(template.Meta.ID),
```
CozeLoop, line 787: [Should-Fix] `TemplateToSubmitExperimentRequest` assumes `template.Meta` is always non-nil

In this helper we dereference `template.Meta` without any guard:

```go
req := &expt.SubmitExperimentRequest{
    WorkspaceID:    workspaceID,
    Name:           gptr.Of(name),
    ExptTemplateID: gptr.Of(template.Meta.ID),
}
...
if template.Meta != nil {
    req.ExptType = gptr.Of(domain_expt.ExptType(template.Meta.ExptType))
}
```

If a corrupted template (or a partially initialised struct in tests) ever has `Meta == nil`, this will panic and `SubmitExptFromTemplateOApi` will return 500 instead of a controlled error.

Since `SubmitExptFromTemplateOApi` already treats `nil` from this helper as an internal error:

```go
submitReq := experiment_convertor.TemplateToSubmitExperimentRequest(template, name, req.GetWorkspaceID())
if submitReq == nil {
    return nil, errorx.NewByCode(errno.CommonInternalErrorCode, errorx.WithExtraMsg("failed to build submit request from template"))
}
```

we can defensively guard `Meta` here and keep the behaviour consistent:

```go
func TemplateToSubmitExperimentRequest(template *entity.ExptTemplate, name string, workspaceID int64) *expt.SubmitExperimentRequest {
    if template == nil || template.Meta == nil {
        return nil
    }

    req := &expt.SubmitExperimentRequest{
        WorkspaceID:    workspaceID,
        Name:           gptr.Of(name),
        ExptTemplateID: gptr.Of(template.Meta.ID),
    }

    if template.TripleConfig == nil {
        return req
    }

    // ... keep the rest of the current field mapping logic unchanged ...
    return req
}
```

This avoids panics while keeping `SubmitExptFromTemplateOApi`’s error handling semantics unchanged.

Priority: Should-Fix

CozeLoop (changes requested):
High-level summary:

Thanks for the comprehensive OpenAPI evaluator/evaluation work. The overall layering (IDL → convertors → application → domain services) is consistent with the existing evaluation module and the new endpoints are wired into the router/handlers in a clean, DDD-friendly way.

However, I found one compatibility-breaking issue and a few behavioural edge cases that should be addressed before merging:

1. **[Must-Fix] Thrift/IDL compatibility**
   - `domain_openapi.evaluator.thrift` changes existing field IDs in `struct Evaluator` (inserting `workspace_id` as field 2 and shifting other fields). This breaks the wire format and violates the IDL compatibility rules. Please keep existing IDs stable and add `workspace_id` with a new ID.

2. **[Should-Fix] OpenAPI input validation and error mapping**
   - `OpenAPIRateLimitDTO2DO` bubbles up `time.ParseDuration` errors directly, and the OpenAPI layer wraps them as generic internal errors. Invalid `rate_limit.period` values should be mapped to a well-defined invalid-parameter error (4xx) instead of 500.
   - `OpenAPIExptTemplateFilterDTO2DO` currently returns `nil` when `logic_op != "and"`, silently discarding both filter conditions and the fuzzy-name keyword. This makes some client inputs appear to be accepted while actually ignored.

3. **[Should-Fix] Template-based experiment submission robustness**
   - `TemplateToSubmitExperimentRequest` assumes `template.Meta` is always non-nil and will panic if that invariant is violated. Given it’s used by `SubmitExptFromTemplateOApi`, it’s safer to guard `Meta` and return `nil` so the caller can surface a controlled error.

I’ve left detailed line comments with concrete suggestions and example code for each issue. Once these are addressed, the PR should be in good shape.

dsf86 (approved):

HymanShi (approved):

PR #431, opened 2026-02-10
[fix][frontend] fix echo bug in trace filter's number input

xzboss:
#### More detailed description for this PR
Fixed the issue where the numeric filter input in the trace list filter would re-display the value after being cleared and losing focus.
tracefilterbug


Maidang1:
Thanks for your contribution.

Maidang1 (dismissed):

mocayo (dismissed):

xzboss:
sorry I forgot to change the user.name...

PR #510, opened 2026-04-29
[feat][evaluation] evaluator record without omitted content

lsy357:
#### What type of PR is this?

fix

#### Check the PR title

- [x] This PR title match the format: \[\<type\>\]\[\<scope\>\] \<description\>. For example: \[fix\]\[backend\] flaky fix

en:
zh(optional):

### DIFF
--- a/backend/modules/evaluation/infra/mq/rocket/producer/expt_event_pub.go
+++ b/backend/modules/evaluation/infra/mq/rocket/producer/expt_event_pub.go
@@ -67,6 +67,10 @@ func newExptEventPublisher(ctx context.Context, cfgFactory conf.IConfigLoaderFac
 			return nil, err
 		}
 
+		if gptr.Indirect(p.cfg.DisableProduce) {
+			continue
+		}
+
 		if !p.cfg.Valid() {
 			return nil, fmt.Errorf("rmq config with invalid addr, key: %v, conf: %v", key, json.Jsonify(p.cfg))
 		}
--- a/backend/modules/evaluation/infra/repo/evaluator/evaluator_record_impl.go
+++ b/backend/modules/evaluation/infra/repo/evaluator/evaluator_record_impl.go
@@ -53,7 +53,11 @@ func (r *EvaluatorRecordRepoImpl) CorrectEvaluatorRecord(ctx context.Context, ev
 	return r.evaluatorRecordDao.UpdateEvaluatorRecord(ctx, po)
 }
 
-func (r *EvaluatorRecordRepoImpl) GetEvaluatorRecord(ctx context.Context, evaluatorRecordID int64, includeDeleted bool) (*entity.EvaluatorRecord, error) {
+func (r *EvaluatorRecordRepoImpl) GetEvaluatorRecord(ctx context.Context, evaluatorRecordID int64, includeDeleted bool, opts ...entity.GetEvaluatorRecordOptionFn) (*entity.EvaluatorRecord, error) {
+	opt := &entity.GetEvaluatorRecordOption{}
+	for _, fn := range opts {
+		fn(opt)
+	}
 	po, err := r.evaluatorRecordDao.GetEvaluatorRecord(ctx, evaluatorRecordID, includeDeleted)
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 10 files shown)


backend/modules/evaluation/infra/mq/rocket/producer/expt_event_pub.go
```diff
@@ -67,6 +67,10 @@ func newExptEventPublisher(ctx context.Context, cfgFactory conf.IConfigLoaderFac
 			return nil, err
 		}
 
+		if gptr.Indirect(p.cfg.DisableProduce) {
+			continue
+		}
```
CozeLoop, line 72: **[🚫Must Fix]** `DisableProduce` skips producer registration and may break callers
- **File**: `backend/modules/evaluation/infra/mq/rocket/producer/expt_event_pub.go:70-72`
- **Issue**: When `DisableProduce` is true, the loop `continue`s and leaves `publisher.producers[key]` unset. Any later `Publish*` call for this key will fail with `rmq producer not found` from `batchSend`, turning a config toggle into a runtime error path.
- **Suggestion**: Make `DisableProduce` a controlled no-op instead of “missing producer”. For example, keep an explicit disabled entry and short-circuit in `batchSend` (or register a noop producer).

Example (short-circuit):
```go
p, ok := e.producers[pk]
if !ok || p == nil || p.p == nil {
    logs.CtxWarn(ctx, "rmq producer disabled or not initialized, key=%v", pk)
    return nil
}
```
- **Reference**: https://go.dev/wiki/CodeReviewComments#errors

backend/modules/evaluation/infra/repo/evaluator/evaluator_record_impl.go
```diff
@@ -53,7 +53,11 @@ func (r *EvaluatorRecordRepoImpl) CorrectEvaluatorRecord(ctx context.Context, ev
 	return r.evaluatorRecordDao.UpdateEvaluatorRecord(ctx, po)
 }
 
-func (r *EvaluatorRecordRepoImpl) GetEvaluatorRecord(ctx context.Context, evaluatorRecordID int64, includeDeleted bool) (*entity.EvaluatorRecord, error) {
+func (r *EvaluatorRecordRepoImpl) GetEvaluatorRecord(ctx context.Context, evaluatorRecordID int64, includeDeleted bool, opts ...entity.GetEvaluatorRecordOptionFn) (*entity.EvaluatorRecord, error) {
+	opt := &entity.GetEvaluatorRecordOption{}
+	for _, fn := range opts {
+		fn(opt)
+	}
```
CozeLoop, line 60: **[🚨Should Fix]** Option parsing can panic on nil and is duplicated
- **File**: `backend/modules/evaluation/infra/repo/evaluator/evaluator_record_impl.go:56-60`
- **Issue**: `for _, fn := range opts { fn(opt) }` will panic if a caller passes a `nil` option function (easy to do when composing opts conditionally). The same parsing logic is duplicated in `GetEvaluatorRecord` and `BatchGetEvaluatorRecord`.
- **Suggestion**: Add a nil-guard and extract a helper constructor to keep the two code paths consistent.

Example:
```go
func NewGetEvaluatorRecordOption(opts ...GetEvaluatorRecordOptionFn) *GetEvaluatorRecordOption {
    o := &GetEvaluatorRecordOption{}
    for _, fn := range opts {
        if fn != nil {
            fn(o)
        }
    }
    return o
}
```
Then use `opt := entity.NewGetEvaluatorRecordOption(opts...)`.
- **Tests**: Add a unit test ensuring `WithoutLoadStorageData()` skips `LoadEvaluatorRecordData` when `recordDataStorage != nil`, and a test that `nil` option does not panic.
- **Reference**: https://go.dev/blog/errors-are-values

CozeLoop (changes requested):
Thanks for the update — the functional-option hook for skipping storage loads makes sense for list/batch scenarios.

However, I think this PR currently introduces a risky runtime failure mode around `DisableProduce` (producer map entry missing => `rmq producer not found` errors). Please address the Must-Fix inline comments first.

Once fixed, please also add targeted unit tests for the new option (`WithoutLoadStorageData`) to prevent regressions.

dsf86 (approved):

xueyizheng (approved):

PR #513, opened 2026-05-07
[feat][evaluation]fix bug 56

tpfz:
#### Check the PR title

en:
zh(optional):

### DIFF
--- a/backend/modules/evaluation/application/experiment_app.go
+++ b/backend/modules/evaluation/application/experiment_app.go
@@ -565,8 +565,71 @@ func (e *experimentApplication) SubmitExperiment(ctx context.Context, req *expt.
 // 2) 从有序 EvaluatorIDVersionList 中批量解析并按输入顺序回填版本ID
 // 3) 从 EvaluatorIDVersionList 中提取 runconfig 和权重配置，构建 evaluator_version_id 到 runconfig/权重的映射
 // 注意：runconfig 用于评估器运行时配置，score weight 用于加权分数计算
+// validateEvaluatorVersionsBelongToWorkspace 校验直接传入的 evaluator_version_id 是否属于当前工作空间
+// 预置评估器（Builtin=true）允许跨空间复用，不做 SpaceID 校验
+func (e *experimentApplication) validateEvaluatorVersionsBelongToWorkspace(ctx context.Context, evaluatorVersionIDs []int64, workspaceID int64) error {
+	if len(evaluatorVersionIDs) == 0 || workspaceID <= 0 {
+		return nil
+	}
+	// 去重，避免重复查询
+	seen := make(map[int64]struct{}, len(evaluatorVersionIDs))
+	uniq := make([]int64, 0, len(evaluatorVersionIDs))
+	for _, id := range evaluatorVersionIDs {
+		if id <= 0 {
+			continue
+		}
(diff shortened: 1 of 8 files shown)


HearyShen (approved):

lsy357 (approved):

backend/modules/evaluation/application/experiment_app.go
```diff
@@ -565,8 +565,71 @@ func (e *experimentApplication) SubmitExperiment(ctx context.Context, req *expt.
...
+			)
+		}
+		// 预置评估器允许跨空间复用
+		if ev.Builtin {
+			continue
+		}
+		if ev.GetSpaceID() != workspaceID {
+			return errorx.NewByCode(
+				errno.EvaluatorVersionNotFoundCode,
+				errorx.WithExtraMsg(fmt.Sprintf("evaluator %d version %s does not belong to workspace %d", ev.ID, ev.GetVersion(), workspaceID)),
```
CozeLoop, line 616: **[🚨SHOULD FIX]** Avoid leaking workspace info in a NotFound-style error
- **File**: `backend/modules/evaluation/application/experiment_app.go:616`
- **Issue**: This returns `EvaluatorVersionNotFoundCode`, but the message explicitly says the evaluator "does not belong to workspace %d". If the error message is surfaced to end users, it weakens the "not found" semantics and may leak cross-workspace authorization details.
- **Suggestion**: Keep the public message generic (e.g. "evaluator version not found") and log the detailed workspace mismatch separately. For example:
```go
logs.CtxWarn(ctx, "evaluator version workspace mismatch: ev=%d ver=%s workspace=%d", ev.ID, ev.GetVersion(), workspaceID)
return errorx.NewByCode(errno.EvaluatorVersionNotFoundCode, errorx.WithExtraMsg("evaluator version not found"))
```

backend/modules/evaluation/application/experiment_app.go
```diff
@@ -624,6 +688,16 @@ func (e *experimentApplication) resolveEvaluatorVersionIDsFromCreateReq(ctx cont
 			if ev == nil {
 				continue
 			}
+			// 非预置评估器必须与实验 WorkspaceID 一致，防止绑定其他空间的评估器
+			// 同时校验根字段 SpaceID（来自 evaluator 元信息）和内层版本 SpaceID（来自 evaluator_version）
```
CozeLoop, line 692: **[💡SUGGESTION]** Keep the SpaceID validation comment consistent with the actual check
- **File**: `backend/modules/evaluation/application/experiment_app.go:692`
- **Issue**: The comment says it validates both the "root SpaceID (evaluator meta)" and the "inner version SpaceID", but the code only checks `ev.GetSpaceID()`, which in `entity.Evaluator` is derived from the version struct. This mismatch can mislead future maintainers.
- **Suggestion**: Update the comment to match the current behavior, or add an explicit meta-level SpaceID check if that is a real requirement.

CozeLoop (changes requested):
Requested changes based on two findings: (1) the new cross-workspace validation returns a NotFound-style code but includes a workspace-mismatch message that may leak authorization details if surfaced to end users; consider making the public message generic and logging details instead; (2) a comment claims both evaluator meta SpaceID and version SpaceID are validated, but the code currently only checks `ev.GetSpaceID()` (version-level).

PR #532, opened 2026-05-25
[fix][evaluation] persist records after eval target timeout

caijialin0626:
## Summary

Fix evaluation failure persistence when the parent execution context is already canceled.

- Persist eval target records with a short detached context derived via `context.WithoutCancel(ctx)`, so `target_record` can still be written after an invoke timeout or parent context cancellation.
- Persist turn and item run logs with the same detached short-timeout pattern, so timed-out synchronous target execution is reflected as failed instead of staying in processing.
- Keep target record diagnostics available through the persisted record: `logid`, `trace_id`, and converted run error.
- Remove temporary Coze video timeout reproduction hooks and E2E marker logs from the final code.

## Root Cause

The synchronous target execution path used the same parent context for both invoking the target and writing the resulting failure records. When the parent context was canceled before persistence, record creation and item/turn log updates could be skipped or fail through canceled context propagation, leaving experiment items stuck in `processing` without a target record log ID.

## Validation

- `go test ./modules/evaluation/domain/service -run 'TestEvalTargetServiceImpl_ExecuteTarget_PersistsFailRecordAfterContextCanceled|Test_ExptItemEvalCtxExecutor_(CompleteSetItemRun|storeTurnRunResult)' -count=1`
- `go test ./modules/evaluation/domain/service -count=1`
- Reproduced the stuck-processing case with a temporary 100ms outer timeout in PPE.
- Verified the fix in PPE: experiment `7590096110691533570` failed with persisted `target_record`, `logid`, `trace_id`, and error.
- Ran normal regression experiment `7590096105115854594`, result `success` with target output fields `actual_output` and `trajectory`.

## Deployment Note

The final branch was deployed through Bits dev self-test task `2401388` for CN. `stone.cozeloop.evaluation` succeeded with commercial dependency commit `7cf57812f18ab36d71c191b69160cb68b14c7da3`, which points to this backend pseudo-version.

### DIFF
--- a/backend/modules/evaluation/domain/service/target_impl.go
+++ b/backend/modules/evaluation/domain/service/target_impl.go
@@ -43,8 +43,6 @@ type EvalTargetServiceImpl struct {
 	configer          component.IConfiger
 }
 
-const evalTargetRecordPersistTimeout = 5 * time.Second
-
 func NewEvalTargetServiceImpl(evalTargetRepo repo.IEvalTargetRepo,
 	idgen idgen.IIDGenerator,
 	metric metrics.EvalTargetMetrics,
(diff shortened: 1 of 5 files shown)


backend/modules/evaluation/domain/service/target_impl.go
```diff
@@ -43,6 +43,8 @@ type EvalTargetServiceImpl struct {
 	configer          component.IConfiger
 }
 
+const evalTargetRecordPersistTimeout = 5 * time.Second
```
lvyingparker, line 46: 这里的超时时间是5s么？

caijialin0626: 这个外层ctx超时，走到了新启动的persist 的ctx的时候保存fail记录的超时时间，就是insert的超时时间

lvyingparker (approved):

ming845378603 (approved):

PR #594, opened 2026-07-24
[feat][evaluation] openapi support async evaluator call

lsy357:
#### What type of PR is this?

feat

#### Check the PR title

- [x] This PR title match the format: \[\<type\>\]\[\<scope\>\] \<description\>. For example: \[fix\]\[backend\] flaky fix

en:
zh(optional):


HearyShen (dismissed):

backend/api/handler/coze/loop/apis/eval_open_apiservice.go
```diff
@@ -291,6 +293,22 @@ func GetExperimentResultExportRecordOApi(ctx context.Context, c *app.RequestCont
...
+	err = c.BindAndValidate(&req)
+	if err != nil {
+		c.String(consts.StatusBadRequest, err.Error())
+		return
+	}
+
+	resp := new(openapi0.AsyncRunEvaluatorOApiResponse)
+
+	c.JSON(consts.StatusOK, resp)
+}
```
CozeLoop, line 297: **[🚫 Must-Fix] The `async_run` endpoint is a non-functional stub — it never invokes the application logic**

**Problem**: This handler is still the default `hz`-generated skeleton. It only calls `c.BindAndValidate(&req)` and then returns an empty, freshly-allocated `resp` with `c.JSON(consts.StatusOK, resp)`. It never dispatches to `localEvalOpenAPIClient.AsyncRunEvaluatorOApi`, so the whole feature (auth check, `AsyncRunEvaluator` submission, `SetEvalAsyncCtx`) in `eval_openapi_app.go` is never reached. Every request to `POST /v1/loop/evaluation/evaluators_versions/:evaluator_version_id/async_run` will return an empty `200 OK` with `invoke_id`/`record` always null.

All sibling handlers in this same file follow the `invokeAndRender` pattern, e.g. `RunEvaluatorOApi` (L190-192), `RunBuiltinEvaluatorOApi` (L268-270) and `AsyncDebugEvalTargetOApi` (L326-328).

**Suggestion**: Replace the body with the standard one-liner so the request is routed to the local Kitex client:
```go
// AsyncRunEvaluatorOApi .
// @router /v1/loop/evaluation/evaluators_versions/:evaluator_version_id/async_run [POST]
func AsyncRunEvaluatorOApi(ctx context.Context, c *app.RequestContext) {
	invokeAndRender(ctx, c, localEvalOpenAPIClient.AsyncRunEvaluatorOApi)
}
```
The now-unused `consts`/`openapi0` imports added at the top of the file should also be removed after the fix (they are otherwise dead imports).

backend/modules/evaluation/domain/service/evaluator_callback_dispatcher.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("X-CozeLoop-Timestamp", timestamp)
+	req.Header.Set("X-CozeLoop-Nonce", nonce)
+	req.Header.Set("X-CozeLoop-Signature", signature)
+
+	resp, err := d.httpClient.Do(req)
+	if err != nil {
+		return err
+	}
```
CozeLoop: **[🚨 Should-Fix] Potential SSRF: the server POSTs to a fully user-controlled `callback_url` without any validation**

**Problem**: `callback_url` originates directly from the OpenAPI request body (`AsyncRunEvaluatorOApiRequest.callback_url` → stored in `EvalAsyncCtx.CallbackURL` at `eval_openapi_app.go:2257`) and is later handed to `doPost`, which issues `http.NewRequestWithContext(ctx, http.MethodPost, url, ...)` against it. There is no scheme allowlist, no host/DNS validation and no private/link-local IP filtering. Any authenticated OpenAPI caller can make the backend issue POST requests to arbitrary internal endpoints (e.g. `http://169.254.169.254/...`, `http://127.0.0.1:<port>/...`, cluster-internal services), which is a classic Server-Side Request Forgery vector. The 3s backoff retry loop amplifies it.

**Suggestion**: Validate `callback_url` before dispatch (ideally at intake in `AsyncRunEvaluatorOApi`, and defensively here):
```go
u, err := url.Parse(callbackURL)
if err != nil || (u.Scheme != "http" && u.Scheme != "https") {
    return fmt.Errorf("invalid callback_url")
}
// resolve host and reject loopback / private / link-local / unspecified ranges
```
Consider an allowlist of permitted domains, and use a `net.Dialer.Control` hook (or a hardened `http.Transport`) that rejects connections to non-public IPs, so DNS-rebinding cannot bypass a pre-check. Reference: OWASP SSRF Prevention Cheat Sheet.

backend/modules/evaluation/domain/service/evaluator_callback_dispatcher.go
```diff
@@ -0,0 +1,100 @@
...
+		secret, _ = d.secretProvider.GetSecret(ctx, spaceID)
+	}
+
+	// 同步 POST + 3s 窗口退避重试；每次重试重算签名
+	if rerr := backoff.RetryThreeSeconds(ctx, func() error {
+		timestamp := strconv.FormatInt(time.Now().Unix(), 10)
+		nonce := GenerateNonce()
+		signature := ComputeHMACSHA256(secret, timestamp+"\n"+nonce+"\n")
+		return d.doPost(ctx, callbackURL, body, timestamp, nonce, signature)
+	}); rerr != nil {
```
CozeLoop, line 72: **[🚨 Should-Fix] Callback delivery blocks the report flow, and the per-attempt timeout (5s) is larger than the whole retry budget (3s)**

**Problem 1 (blocking)**: `Dispatch` is called synchronously inside `ReportEvaluatorInvokeResult_` (`eval_openapi_app.go:2832`). The internal report interface therefore blocks on an *external, user-controlled* endpoint for the full retry window. For an async-evaluator feature this couples the internal completion path to third-party latency and can be abused to slow the report interface. The existing `WebhookDispatcher` offloads retries to MQ (`PublishExptWebhookNotifyEvent`) instead of retrying inline.

**Problem 2 (timeout vs budget)**: `httpClient.Timeout = 5s` (L41) but `backoff.RetryThreeSeconds` sets `MaxElapsedTime = 3s`. When the callback endpoint is slow/unresponsive, the first attempt alone (up to 5s) already exceeds the 3s elapsed budget, so `backoff` returns `Stop` and **no retry ever happens** — the retry loop is effectively dead for exactly the timeout case it is meant to cover.

**Suggestion**:
- Dispatch asynchronously (a managed goroutine with `context.Background()` + a bounded worker/semaphore, or reuse the webhook MQ path) so the report interface returns immediately.
- Make the per-attempt timeout strictly smaller than the retry budget (e.g. per-attempt `1s` with `RetryThreeSeconds`, or switch to `RetryWithMaxTimesAndInterval` for a deterministic attempt count).

idl/thrift/coze/loop/evaluation/coze.loop.evaluation.openapi.thrift
```diff
@@ -1048,6 +1048,44 @@ struct RunEvaluatorOpenAPIData {
...
+
+// 异步评估器执行完成后，服务端主动 POST 给 callback_url 的回调 body
+struct EvaluatorCallbackPayloadOApi {
+    1: optional string cid (go.tag = 'json:"cid"')                                          // 本次回调投递的唯一 ID（服务端生成，用于重试去重）
+    2: optional i64 invoke_id (go.tag = 'json:"invoke_id"')                                 // = async_run 返回的 invoke_id
+    3: optional i64 workspace_id (go.tag = 'json:"workspace_id"')
+    4: optional i64 evaluator_version_id (go.tag = 'json:"evaluator_version_id"')
+    5: optional string status (go.tag = 'json:"status"')                                    // success | fail
+    6: optional evaluator.EvaluatorOutputData output (go.tag = 'json:"output,omitempty"')   // 仅 success 时携带
+    7: optional i64 time_consuming_ms (go.tag = 'json:"time_consuming_ms"')
```
CozeLoop: **[🚨 Should-Fix] `i64` fields in the callback payload are serialized as JSON numbers — precision loss for JS consumers and inconsistent with the `async_run` response**

**Problem**: `EvaluatorCallbackPayloadOApi` is marshaled with `pkg/json` (Go `encoding/json`) using only `go.tag`, so `invoke_id`, `workspace_id`, `evaluator_version_id` and `time_consuming_ms` are emitted as raw JSON numbers. IDs beyond 2^53 will lose precision in any JavaScript/JSON consumer. This also contradicts `AsyncRunEvaluatorOpenAPIData.invoke_id` (L1074), which uses `api.js_conv="true"` and is returned to the client as a **string** — so the same `invoke_id` is a string in the `async_run` response but a number in the callback body. The IDL spec's 【JS i64 兼容】 rule requires i64 fields to be JS-safe.

**Suggestion**: Since this struct is not gateway-serialized (`api.js_conv` has no effect on a server-originated POST body), emit the i64 fields as strings via the go.tag `string` option so they match the response representation:
```thrift
2: optional i64 invoke_id            (go.tag = 'json:"invoke_id,string"')
3: optional i64 workspace_id         (go.tag = 'json:"workspace_id,string"')
4: optional i64 evaluator_version_id (go.tag = 'json:"evaluator_version_id,string"')
7: optional i64 time_consuming_ms    (go.tag = 'json:"time_consuming_ms,string"')
```
**Reference**: CozeLoop IDL 规范 §【JS i64 兼容】.

CozeLoop (changes requested):
### 【Aime CR】Async Evaluator Callback Review

Reviewed the non-generated changes of this PR (async evaluator OpenAPI + HMAC callback dispatcher). Generated code (`kitex_gen`, `loop_gen`, `wire_gen.go`, `mocks`, `errno`) and the `reflect.Ptr`→`reflect.Pointer` / `fmt.Sprintf`→`fmt.Fprintf` lint-only cleanups were reviewed and carry no risk.

**Findings**
- 🚫 **1 Must-Fix**: the `async_run` HTTP handler is a leftover hz stub and never calls the application logic — the endpoint is completely non-functional.
- 🚨 **3 Should-Fix**: (1) SSRF via unvalidated user-supplied `callback_url`; (2) synchronous callback blocks the report flow and the 5s per-attempt timeout defeats the 3s retry budget; (3) i64 fields in the callback payload lose JS precision and are inconsistent with the `async_run` response.

The async-ctx key (`evaluator:%d`) correctly matches the reader in `ReportEvaluatorInvokeResult_`, and the `Event == nil` path for the OpenAPI flow is safe (nil-guarded getters). Please address the must-fix before merge. Detailed inline comments below.

HearyShen (dismissed):

xueyizheng (dismissed):

HearyShen (approved):

xueyizheng (approved):

